PR #106, opened 2021-05-07
Test flash via WebUSB.

microbit-matt-hillsdon:
Requires non-headless mode and a device connected.

Need to discuss the auto-connection in device.ts. I think MakeCode does
this already. We could make it a test-only behaviour if desired.


microbit-matt-hillsdon:
Closing for now but leaving branch. I'll rework and submit another PR.

PR #186, opened 2021-07-06
Code structure highlighting: alternating colours

microbit-matt-hillsdon:
Demo only for discussion.


microbit-matt-hillsdon:
I think we're agreed this isn't the way to go. Closing PR.

Issue #207, opened 2021-07-22
[Code structure] Config options for experimentation


microbit-matt-hillsdon:
We've agreed to add a bunch of configuration options to make the code structure highlighting option more configurable.

We'll also add and/or document some CSS variables to enable easy experimenting with the colours used.

- [x] As a first step we should collapse the separate implementations. The l-shaped-boxes one should have all the information needed to draw the others. When we went with separate implementations the expectation was that we were going to make a quick choice but that seems less likely now.
- [x] Add functional options documented in Paper doc to settings
- [x] Add colour options
- [x] Document options for initial Foundation feedback
- [x] Revise/merge

microbit-matt-hillsdon:
This will probably also fix #204 or at least make it trivial.

microbit-matt-hillsdon:
Borders and l-shapes are a problematic combo if we stick with rounded corners

![image](https://user-images.githubusercontent.com/44397098/126797450-9f7c94cd-4905-49d8-b1d3-626cbc445eeb.png)

We might be able to make it work with an SVG but that would require a degree of reworking. Would probably let you round the inner corner of the L though.

microbit-matt-hillsdon:
Hard to say whether this left-edge-only boxes background combo is a good thing either:

![image](https://user-images.githubusercontent.com/44397098/126798242-104cc9af-5b52-47a3-a2b0-6a38e5b4da2f.png)

microbit-matt-hillsdon:
We can't use CSS for hover in a straightforward way as L-shapes are made up of multiple boxes. I guess we might as well use event listeners for both cases for now, potentially simplifying in future.

microbit-matt-hillsdon:
I'll leave this on a branch for an initial round of feedback.

The most significant change for someone who doesn't engage with the options is stopping trimming trailing whitespace from the scope of a block. We could reinstate that but then I think it would be critical that we didn't highlight when we were in that whitespace as the mismatch is jarring.

jaustin:
Playing with the settings on the latest branch made https://review-python-editor-next.microbit.org/code-structure-settings-jonny/ I've realised a couple of things

The approach I took where I made the 'selected block' darker than the block background kinda limits (a) how strong the main block background colours can be and (b) means that we get into contrast issues sooner. For example it's harder to see the structure highlighting of the deeper nested bits of code once you've selected a lightly nested block...

[...]

PR #209, opened 2021-07-23
Code structure highlighting: collapse implementations before adding more config

### LINKED ISSUE #204: Code structure highlighting: missing colours from default theme
Code structure highlighting is missing colours in the default theme. At least the block colour is missing.

microbit-matt-hillsdon:
This removes the "brackets" implementation that we've decided against progressing and collapses the code that implements the remaining approaches into a single parameterised implementation. We lean more heavily on CSS, but there is a fundamental difference between l-shapes a boxes that needs switching in the view code.

The new parameters are not yet exposed to the user. Combinations of parameters that aren't surfaced through the settings will need careful review after this PR.

See #207 

I've been testing this with the default theme and it also fixes #204


microbit-matt-hillsdon:
I'll do some side-by-side visual review before merging with the test file:

- [ ] L-shapes
- [ ] L-shape boxes
- [ ] Boxes

microbit-matt-hillsdon:
I'll do some side-by-side visual review before merging with the test file:

- [ ] L-shapes
- [ ] L-shape boxes
- [ ] Boxes

microbit-matt-hillsdon:
One tweak later (missing color in the theme) and this is looking good visually using https://github.com/microbit-foundation/python-editor-next/blob/main/testData/highlighting.py to compare.

PR #211, opened 2021-07-23
Code structure settings.

microbit-matt-hillsdon:
These are experiments and aren't expected to remain configurable by
users. We'll probably hide them for a beta.

- [x] Reinstate or remove use of doc-utils.

### DIFF
--- a/src/common/use-local-storage.ts
+++ b/src/common/use-local-storage.ts
@@ -17,10 +17,21 @@ export function useLocalStorage<T>(
     const value = localStorage.getItem(key);
     if (value !== null) {
       try {
-        const parsed = JSON.parse(value);
+        let parsed = JSON.parse(value);
+        // Ensure we get new top-level defaults.
+        parsed = {
+          ...defaultValue,
+          ...parsed,
+        };
+        // Remove any top-level keys that aren't present in defaults.
+        const unknownKeys = new Set(Object.keys(parsed));
+        Object.keys(defaultValue).forEach((k) => unknownKeys.delete(k));
+        unknownKeys.forEach((k) => delete parsed[k]);
+
         if (!validate(parsed)) {
           throw new Error("Invalid data stored in local storage");
         }
+
         return parsed;
       } catch (e) {
         // Better than exploding forever.
--- a/src/settings/SettingsArea.tsx
+++ b/src/settings/SettingsArea.tsx
@@ -120,6 +120,24 @@ const SettingsArea = () => {
           }
         />
         <SelectFormControl
+          id="codeStructureBackground"
+          label="Background"
+          options={[
+            {
+              value: "none",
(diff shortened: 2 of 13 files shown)


src/common/use-local-storage.ts
```diff
@@ -17,10 +17,21 @@ export function useLocalStorage<T>(
...
+        // Ensure we get new top-level defaults.
+        parsed = {
+          ...defaultValue,
+          ...parsed,
+        };
+        // Remove any top-level keys that aren't present in defaults.
+        const unknownKeys = new Set(Object.keys(parsed));
+        Object.keys(defaultValue).forEach((k) => unknownKeys.delete(k));
+        unknownKeys.forEach((k) => delete parsed[k]);
+
```
microbit-matt-hillsdon, line 30: We should test this or perhaps swap it out for something that's solved this problem well (though it could easily be excessive code).

src/settings/SettingsArea.tsx
```diff
@@ -61,68 +64,195 @@ const SettingsArea = () => {
...
+
+        <BooleanFormControl
+          id="codeStructureHoverBorder"
+          label="Highlight border on mouse over"
+          value={settings.codeStructureHighlighting.hoverBorder || false}
+          onChange={(hoverBorder) =>
+            setSettings(modifyCodeStructureSettings(settings, { hoverBorder }))
+          }
+        />
+      </VStack>
```
microbit-matt-hillsdon, line 182: These settings aren't implemented yet. I'll do that on Monday. 

Should just be adding few classes and faffing with the CSS for hover.

Cursor position will require us to add/update a class.

src/settings/SettingsArea.tsx
```diff
@@ -61,68 +64,195 @@ const SettingsArea = () => {
...
+          label="Shape"
+          options={[
+            {
+              value: "l-shape",
+              label: "L-shapes",
+            },
+            {
+              value: "box",
+              label: "Boxes",
+            },
```
microbit-matt-hillsdon, line 115: The text here and throughout could be extracted for i18n but not sure it's worth it for these experimental settings. The "Common" heading maybe should be, once we're happy with the text?

src/common/use-local-storage.ts
microbit-matt-hillsdon: Test added in 3a4eb4b655522fb796c8ef6a3d5891614708b592.

microbit-matt-hillsdon:
Code changes look good but this needs more feedback / practical testing before merge.

microbit-matt-hillsdon:
Closing in favour of #249 

I'll leave the branch around for a while though as it's useful for experimentation.

We might also want to bring back the settings UI as an advanced option if more configurability is needed.

PR #213, opened 2021-07-27
Updated indent behaviour

microbit-matt-hillsdon:
Inlines https://github.com/microbit-matt-hillsdon/lang-python/tree/indent-rethink for the moment for testing.

Hopefully there'll be upstream interest in something similar to this rather than maintaining it this way.

We could also consider publishing a package. Depending on a GitHub branch doesn't seem to play well with installing the private theme package for reasons I don't really fancy debugging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13393,6 +13393,14 @@
         "lezer-tree": "^0.13.2"
       }
     },
+    "lezer-python": {
+      "version": "0.13.7",
+      "resolved": "https://registry.npmjs.org/lezer-python/-/lezer-python-0.13.7.tgz",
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
The most frustrating behaviours have been fixed upstream and we've upgraded. There are still differences between a body at the end of the document and one elsewhere and various minor enhancements from my PR would be nice to have too. Need to find the time to go through and raise more specific issues.

microbit-matt-hillsdon:
Closing for now. We'll revisit after user testing and before beta as now isn't the time for risky change

PR #214, opened 2021-07-28
Errors: TigerPython and runtime errors

microbit-matt-hillsdon:
Tackling these together as I think we need to agree on differing UX for the lint and runtime errors.

We might end up feature flagging parts of this PR though.

## Runtime errors

Captures the last traceback error from the last 4k of serial output. Discards when serial is reset.

Shown in a new, default, compact form of the serial area. We could also consider displaying the message line.

Basic navigation to the file / line number that causes the error from the summary in the compact view and also hyperlinked each line of the traceback.

Interesting to consider that it can be out of sync with the code, e.g. you've never flashed but you have an error from an old program.

No current indication of the line other than the hyperlink. Definitely a weak point of this UX. We could add a gutter marker or try a completely different presentation of the error, e.g. one like this: https://codemirror.net/demo/widget.html (but less scary, and only the one runtime error).

## TygerPython

Runs in a web worker.

Just uses CM's default rendering for now (though using our fonts).

We can align the colours to our error and success colours if that seems like a good move.

F8 and Ctrl/Cmd-Shift-M trigger built-in CM behaviours that we can choose to keep, style, disable or reimplement. I think we need something like the Cmd-Shift-M behaviour for accessibility of lint messages.

TygerPython just gives us a line/column. We currently underline from that column to the end of the line. This doesn't amount to much when the error is at the end of the line and the message is barely visible. This is quite common for missing colon/bracket/quote errors. We could adapt TygerPython's output to ensure we highlight at least the last token on the line? Other ideas welcome.

It's been noted that the errors could be clearer in some cases (e.g. "unterminated"). [Full set of messages](https://github.com/Tobias-Kohn/TigerPython-Parser/blob/master/tpParser/shared/src/main/scala/tigerpython/parser/errormessages/EnglishMessages.scala). I suspect only the most common ones are worth investing any effort in, though I appreciate that’s hard to judge. There are a few that won’t apply to us (for Python 2.x or disabled features in support of some non-standard constructs for teaching, e.g. repeat loop). Easy to override a few of them, at least until we consider translations.

TigerPython does not attempt type checking so doesn't know that e.g. disply.scroll("Foo") is an error.

https://github.com/microsoft/pyright is an option we can investigate (much more work, may not be possible to use at all, likely to have quite technical messages).

## Todo

- [ ] Demos, feedback, next steps. Make sure we discuss the design UI.
- [ ] Figure out a good way to e2e test this without the device. We need to start stubbing device interactions.
- [ ] Consider whether we should announce traceback errors for screenreaders. Do we have a go to line action?


src/device/device-hooks.test.tsx
```diff
@@ -0,0 +1,104 @@
+import { TracebackScrollback } from "./device-hooks";
+
+const toCrLf = (text: string): string =>
+  text.replace(/[\r\n]/g, "\n").replace(/\n/g, "\r\n");
+
+describe("TracebackScrollback", () => {
```
microbit-matt-hillsdon, line 6: Need to augment these tests for the Traceback object's line / file properties.

src/device/device-hooks.ts
```diff
@@ -45,3 +48,124 @@ export const useConnectionStatus = () => {
...
+        return file;
+    }
+  }
+}
+
+export class TracebackScrollback {
+  private scrollback: string = "";
+  push(data: string) {
+    this.scrollback = this.scrollback + data;
+    const limit = 4096;
```
microbit-matt-hillsdon, line 104: We should test this limit behaviour.

src/editor/codemirror/linting/linter.ts
```diff
@@ -0,0 +1,37 @@
...
+        (e: MessageEvent) => {
+          const errors = e.data;
+          resolve(
+            errors.map((error: any) => {
+              return {
+                severity: "error",
+                // So that the go-to-next error feature works we highlight
+                // from the offset to the end of the line (as we don't have
+                // an end position). Not great when the error is at the end
+                // of a line.
```
microbit-matt-hillsdon, line 23: It's really quite bad at the end of the line and errors there are common (missing quote, colon, bracket etc).

src/editor/codemirror/linting/linter.worker.ts
```diff
@@ -0,0 +1,19 @@
+import * as TPyParser from "tigerpython-parser";
+
+TPyParser.setLanguage("en");
+TPyParser.rejectDeadCode = true;
+TPyParser.strictCode = true;
+
+/* eslint-disable no-restricted-globals */
+let workerSelf = self as unknown as BroadcastChannel;
```
microbit-matt-hillsdon, line 8: I don't think this is really the right interface but it had the correct definition of postMessage. Is there something saner I can do short of splitting this out into a separate project?

src/serial/SerialArea.tsx
```diff
@@ -19,9 +43,124 @@ const SerialArea = (props: BoxProps) => {
...
+          height="100%"
+        >
+          <SerialBar
+            height={12}
+            compact={compact}
+            onSizeChange={onSizeChange}
+          />
+          <XTerm
+            visibility={compact ? "hidden" : undefined}
+            height="calc(100% - 40px)"
```
microbit-matt-hillsdon, line 60: This 40px is almost certainly wrong. Value should be shared between here and the split view.

src/serial/SerialArea.tsx
```diff
@@ -19,9 +43,124 @@ const SerialArea = (props: BoxProps) => {
...
+    },
+    [setSelection, traceback]
+  );
+  return (
+    <Link textDecoration="underline" onClick={handleClick}>
+      {children}
+    </Link>
+  );
+};
+
```
microbit-matt-hillsdon, line 165: Probably worth splitting out.

src/device/device-hooks.ts
microbit-matt-hillsdon: 43d7b55107b63ca60d0273d2dd328559bff7f99d

src/editor/codemirror/linting/linter.worker.ts
microbit-matt-hillsdon: 2813f39940c9e9dae986ae41e4787fd2b5f5ec09

src/serial/SerialArea.tsx
microbit-matt-hillsdon: 9265c8cdc1ae08b812e6f3e0debe2743fb0a503b

src/editor/codemirror/linting/linter.ts
microbit-matt-hillsdon: One for discussion rather than a quick tech fix I think. Moved to the PR summary.

microbit-matt-hillsdon:
I'm going to close this PR as we've decided to focus on what we can get out of Pyright (see prototyping on #222).

Always possible we'll revisit so I'll keep the branch around. If we do pick it up then we will need to disentangle the runtime errors work. That work has been merged separately as #224.

PR #216, opened 2021-07-31
proof of concept for downloading a file from microfs

pdg137:
This is a proof of concept for accessing the [microfs filesystem](https://github.com/ntoll/microfs/) on the micro:bit. This is the only non-volatile data storage option for the micro:bit, so it seems like we should offer a way to access it.

Note that I don't know what I'm doing with Node/Typescript/React at all and have just barely gotten it to work. So this is definitely a draft request! But I thought it was worth putting it up here in case anyone has feedback.

* Is this something the maintainers would be interested in including as a feature?
* Any general comments on my approach?
* Specific suggestions about how to structure the code?

## How to use it

Here's an example program that will write 100 lines to `data.txt`:

```python
from microbit import *
from os import *
import random

file = open("data.txt", "w")
for i in range(100):
    file.write(str(random.randint(1,100))+" hello!!\r\n")
file.close()

display.show("!")
```

The data download button is here:

image

Currently it just looks for a file with that name and downloads it.  There's no error handling, for example if the file does not exist, and no timeouts at all.  Assuming all goes well, your browser should pop up a download dialog or just save `data.txt` to your Downloads folder.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -175,6 +175,10 @@
     "defaultMessage": "Download a project hex file",
     "description": "Hover text over download button"
   },
+  "download-microfs-files": {
+    "defaultMessage": "Download MicroFS files",
+    "description": "Download button menu option for downloading on-board MicroFS files"
+  },
   "download-python": {
     "defaultMessage": "Download Python script",
     "description": "Download button menu option for downloading Python script"
--- a/src/messages/en.json
+++ b/src/messages/en.json
@@ -355,6 +355,12 @@
       "value": "Download a project hex file"
     }
   ],
+  "download-microfs-files": [
+    {
+      "type": 0,
+      "value": "Download MicroFS files"
+    }
+  ],
   "download-python": [
     {
       "type": 0,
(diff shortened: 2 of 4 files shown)


microbit-matt-hillsdon:
This is interesting! Just a heads up that most of the folks at micro:bit who I'd discuss new editor features with are on summer holidays at the moment so it might be a while before I get a chance to properly discuss this.

I worry a bit that there's potential for this to confuse some users as there's already a list of Files in the app and various ways to download/flash but it's not quite the same thing as the live file system from the device. It might be hard to make that distinction clear enough for students.

Although not directly related, we've recently added some data logging support for micro:bit V2 devices that have more flash. This is still in testing and we're ironing out bugs, but you can play with it in MakeCode today and it'll be coming to MicroPython too:

https://microbit.org/get-started/user-guide/data-logging/

If you haven't already, you can also find sign-up details for our tech newsletter and Slack on https://tech.microbit.org/

pdg137:
I agree about the download button being confusing like this. But in fact, the whole concept of the microfs being different from the mass storage filesystem is confusing. With this new data logging system (which looks great!) coming to MicroPython, is there any chance that we'll be able to interact more generically with the mass storage files with `open()`, `write()`, and so on?

Anyway, thanks for the thoughtful comments. I signed up for the mailing and applied for the Slack channel and maybe can continue the discussion more informally over there.

Issue #215, opened 2021-07-31
[Serial] Pasting several lines into serial terminal causes "Bad status 8 -> 255" error and disconnect


pdg137:
## Bug Description

Pasting the following script into the terminal causes an instant disconnection:

```python
print("hello, world")
print("hello, world")
print("hello, world")
print("hello, world")
print("hello, world")
```
## Screenshots

![image](https://user-images.githubusercontent.com/466760/127748125-4ec20e92-f2d8-4f14-aaa3-cefbf32d9365.png)

## Environment

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Chrome
 - Latest: Version 92.0.4515.107 (Official Build) (64-bit)

## Additional context

Possibly related: #89.

I discovered this when writing long strings from Javascript using `device.serialWrite()`. My goal is to implement [microfs](https://github.com/ntoll/microfs) file downloads for getting data off of the micro:bit.

microbit-matt-hillsdon:
Reproduces for me on Mac.

I wanted to see if I could reproduce this on https://python.microbit.org/v/2 (as the serial code is very similar but the terminal is different) but there seems to be a short limit on the amount of pasted content there. It's not obviously something the v2 hterm setup is configuring though (Shift-Ins is paste there).

DETAILS.TXT from the micro:bit I reproduced with:

```
# DAPLink Firmware - see https://mbed.com/daplink
Unique ID: 9904360251974e450039900a00000041000000009796990b
HIC ID: 9796990b
Auto Reset: 1
Automation allowed: 0
Overflow detection: 0
Incompatible image detection: 1
Page erasing: 0
Daplink Mode: Interface
Interface Version: 0255
Bootloader Version: 0254
Git SHA: 1436bdcc67029fdfc0ff03b73e12045bb6a9f272
Local Mods: 0
USB Interfaces: MSD, CDC, HID, WebUSB
Bootloader CRC: 0xe5ec290c
Interface CRC: 0x5b5cc0f5
Remount count: 0
URL: https://microbit.org/device/?id=9904&v=0255
```

More investigation required.

microbit-matt-hillsdon:
Pasting 63 characters, e.g. `012345678901234567890123456789012345678901234567890123456789012` results in the disconnect. Pasting 62 characters is OK for me. It doesn't seem to be related to lines as such.

[...]

PR #218, opened 2021-08-02
Send serial writes in max 62 byte chunks.

microbit-matt-hillsdon:
I'll probably leave this one for a bit to discuss with someone who's been more involved in this sort of WebUSB issue.

See https://github.com/microbit-foundation/python-editor-next/issues/215

### DIFF
--- a/src/device/device.ts
+++ b/src/device/device.ts
@@ -431,7 +431,16 @@ export class MicrobitWebUSBConnection extends EventEmitter {
   serialWrite(data: string): Promise<void> {
     return this.withEnrichedErrors(async () => {
       if (this.connection) {
-        return this.connection.daplink.serialWrite(data);
+        // Using WebUSB/DAPJs we're limited to 64 byte packet size with a two byte header.
+        // https://github.com/microbit-foundation/python-editor-next/issues/215
+        const maxSerialWrite = 62;
+        let start = 0;
+        while (start < data.length) {
+          const end = Math.min(start + maxSerialWrite, data.length);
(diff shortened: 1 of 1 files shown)


src/device/device.ts
```diff
@@ -431,7 +431,16 @@ export class MicrobitWebUSBConnection extends EventEmitter {
   serialWrite(data: string): Promise<void> {
     return this.withEnrichedErrors(async () => {
       if (this.connection) {
-        return this.connection.daplink.serialWrite(data);
+        // Using WebUSB/DAPJs we're limited to 64 byte packet size with a two byte header.
+        // https://github.com/microbit-foundation/python-editor-next/issues/215
```
microbit-matt-hillsdon, line 435: Should this be in DAPJs or at least documented on serialWrite?

microbit-matt-hillsdon:
@microbit-carlos does this seem reasonable to you? See the issue for reproductions which also apply to the V2 editor.

If so, do you think there's a case for the fix being at a lower level in DAPJs? Wouldn't the limit apply to all uses?

src/device/device.ts
microbit-matt-hillsdon: Chatted to @microbit-carlos and did a bit more digging.

I think if we were using the node-usb DAPJs transport then this might have been addressed automatically (by libusb). Not had a chance to try yet. See docs at https://github.com/tessel/node-usb#outendpoint

So I think it's worth following up and seeing if the fix should be in the WebUSB transport. Not going to hold this up on that though, but I'll reopen the issue for follow up.

Issue #217, opened 2021-08-01
In serial terminal, the keyboard shortcuts Ctrl+C and Ctrl+D are broken or confusing


pdg137:
## Bug Description

The following keyboard shortcuts have issues in the xterm / serial terminal / SerialArea:

* Even though Ctrl+V works for paste, Ctrl+C does not copy text.  There's also no option to copy in the right-click menu. Ctrl+C doesn't even copy the selected text if you click out of the xterm control first. So even though you can select, there appears to be no way to copy.
* Ctrl+D does a "soft reboot" but confusingly also creates a browser bookmark.

## How To Reproduce

1. Connect to a micro:bit.
2. Try Ctrl+C and Ctrl+D in the serial terminal area.

## Expected behavior

* An alternate key like Ctrl+Shift+C should be available for copy and mentioned in the introductory text along with Ctrl+C.
* There should be user-interface widgets such as buttons or a right-click menu to support copy and paste.
* The Ctrl+D shortcut should block the bookmark feature of Chrome.

## Screenshots

Can select text but no way to copy:

![image](https://user-images.githubusercontent.com/466760/127761338-72a976fc-befa-4e22-9871-4cea3b97e29d.png)

Annoying bookmark-and-soft-reboot shortcut:

![image](https://user-images.githubusercontent.com/466760/127761392-c936b609-b551-4615-85c2-fe12155c243c.png)
## Environment

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Chrome
 - Version 92.0.4515.107

microbit-matt-hillsdon:
Thanks for the great report. I'll look at this one later in the week.

Ctrl-C will need to remain unbound in order to interrupt a running micro:bit program. Ctrl-Shift-C/V is a common solution but I'll look at others and have a think about making this discoverable. https://python.microbit.org/v/2 copies text on selection and shows an icon indicating it has done so, which might be surprising for some but is at least very discoverable. That is default hterm behaviour (v2 uses hterm, the alpha uses xterm).

I've used Ctrl-D lots, but on Mac mostly, so I'd missed that it also creates a bookmark on Windows.

Useful reference: https://github.com/xtermjs/xterm.js/issues/2478

microbit-matt-hillsdon:
Fixed the triggering of the bookmark action in #219 and added Ctrl-Shift+C for copy in the same PR.

[...]

PR #219, opened 2021-08-02
First stab at configuring xterm.js keys.

microbit-matt-hillsdon:
Attempt to avoid clash with Ctrl-D browser shortcut on Windows.

Provide alternative copy shortcut (Mac works fine with Cmd+C but Windows
has a clash with Ctrl-C).

I think paste is working via a paste event sent via the browser and we
can't trigger it ourselves for browser security reasons. On Mac Cmd-V
and Shift-Insert work. Yet to test this on Windows.

TODO:
- [x] Test on Windows
- Discuss more discoverable UX, perhaps after merging a serial toolbar / status bar.
    - Moved to https://github.com/microbit-foundation/python-editor-next/issues/220

Closes https://github.com/microbit-foundation/python-editor-next/issues/217

### DIFF
--- a/src/serial/XTerm.tsx
+++ b/src/serial/XTerm.tsx
@@ -18,6 +18,7 @@ import {
 import { EVENT_SERIAL_DATA, EVENT_SERIAL_RESET } from "../device/device";
 import { useDevice } from "../device/device-hooks";
 import "./xterm-custom.css";
+import customKeyEventHandler from "./xterm-keyboard";
 
 const ptToPixelRatio = 96 / 72;
 
@@ -46,6 +47,7 @@ const XTerm = (props: BoxProps) => {
       });
       const fitAddon = new FitAddon();
       term.loadAddon(fitAddon);
+      term.attachCustomKeyEventHandler(customKeyEventHandler);
       term.open(ref.current);
 
       let firstWrite = true;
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
TODO:
- [ ] Test on Windows
- [ ] Discuss more discoverable UX, perhaps after we've merged some kinda of serial toolbar or status bar.

microbit-matt-hillsdon:
Tried it on Windows.

The PR fixes the bookmark issue.

On Windows 10 the new Ctrl-Shift+C shortcut works. Ctrl+V works as does (by luck) Ctrl-Shift+V for paste, presumably because the shift isn't taken into account.

Comparing to v2, I haven't really understood the logic hterm uses but its defaults seem to work well for us there and its not affected by this bug. Ctrl-P and Ctrl-L browser shortcuts don't work but Ctrl-N and Ctrl-T do for example. It'd be better to retain those where we can in the alpha for accessibility as I doubt folks are writing terminal apps with shortcuts on the micro:bit. Ctrl-Shift-V works for paste and Ctrl-V does not.

I think we should suppress propagation of all the Ctrl key combos that have meaning to MicroPython as e.g. Ctrl-E for paste mode also focuses the address bar in for a Google search and there's no sense reviewing all of them in all browsers.

PR #222, opened 2021-08-05
LS based errors

microbit-matt-hillsdon:
Very much a work in progress. For discussion/demos only at this point.

## Demo prep
- [x] Disentangle the runtime errors work from this (and #214) and merge that separately.
    - Now https://github.com/microbit-foundation/python-editor-next/pull/224. This needs rebasing on main once that's merged.
- [x] Remove TygerPython cruft from this PR. I expected to fail quickly but instead I've ended up with a successful mess!

## Investigation

- How viable is it to override common error messages to improve them? They're in resource bundles etc. but how general are the error codes?
    - The type-related messages are pretty specific.
    - The syntax related ones are dominated by 40+ expectedFoo message codes, e.g. `expectedSliceIndex` and a few lower-level parse issues.
    - I think we need to experiment with this with an idea of the messages we care about.

## Minimal mergeable PR

- [ ] Sort out the language server client 
    - I think we should replace with something based on vscode-jsonrpc that supports multiple files and perhaps workspaces from the start. It can drop features, we can add them back incrementally. This means it needs an API beyond just being a CM extension and integration with the rest of the app (file create, delete, open, close, edit). We should also do incremental edit support. 
    -  It would be good to improve typeshed initialization. Can we make it part of the initialize message? 
    - We should probably open all Python files in the project (there won't be many). That way we'll own them and the language server won't expect them to be in the file system.
    - We can tell the language server when file operations occur. Not sure if Pyright supports this 
    - We should remove (non-error) hovers for now

- [ ] Acceptable quality stubs
    - Matt to chat to Carlos about approach

- Sufficient feature quality (to be discussed) for:
    - [ ] error messages (but no text customisation)
    - [ ] autocomplete (can be split if sensible)

- [ ] Licensing. As the built code is in our repo we should include really clear licensing info with a copy of Microsoft's license. I think ideally we'd build it and package it but not sure we're ready for that quite yet.
 
 - [ ] Test Pyright with Monaco, perhaps inside the Pyright fork. A testing environment for Pyright in the browser should be using Monaco to uncover problems rather than a minimal language server that's also undergoing change.

## Later

- [ ] Customise common error messages in educational scenarios.
- [ ] Improve stub docs and discuss exposing hover information. The hover UX might not be what we want but the information has value. Might need lots of edu team info on quality of text for code stubs though.
- [ ] Parameter hovers. VS Code shows completion when you're between the brackets. It shows the current parameter in bold etc.


craco.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     configure: {
       moduleNameMapper: {
         "^@deployment(.*)$": `<rootDir>/${location}$1`,
+        "\\.worker": "<rootDir>/src/mocks/worker.js",
```
microbit-matt-hillsdon, line 25: Not needed for this approach.

package.json
```diff
@@ -39,8 +40,11 @@
     "react-dom": "^17.0.2",
     "react-icons": "^4.2.0",
     "react-intl": "^5.20.4",
+    "tigerpython-parser": "file:./third-party/tigerpython-parser",
```
microbit-matt-hillsdon, line 43: Not needed here.

package.json
```diff
@@ -39,8 +40,11 @@
...
+    "vscode-languageserver-protocol": "^3.16.0",
     "web-vitals": "^1.1.1",
     "word-wrap": "^1.2.3",
+    "worker-loader": "^3.0.8",
```
microbit-matt-hillsdon, line 47: Not needed here.

package.json
```diff
@@ -78,6 +82,9 @@
     "i18n:compile": "node bin/sort.js && formatjs compile-folder --ast lang src/messages"
   },
   "eslintConfig": {
+    "ignorePatterns": [
+      "src/third-party/**"
+    ],
```
microbit-matt-hillsdon, line 87: Not needed.

src/editor/codemirror/config.ts
```diff
@@ -53,9 +53,9 @@ export const editorConfig: Extension = [
   bracketMatching(),
   closeBrackets(),
   highlightStyle(),
-  autocompletion({
-    override: completion,
-  }),
+  //autocompletion({
+  //  override: completion,
+  //}),
```
microbit-matt-hillsdon, line 58: Tidy. If we're happy with the approach we can kill the stub autocomplete code.

src/editor/codemirror/config.ts
```diff
@@ -74,4 +74,7 @@ export const editorConfig: Extension = [
   EditorState.tabSize.of(indentSize), // But hopefully not used!
   indentUnit.of(" ".repeat(indentSize)),
   python(),
+  lspLinter(),
+  autocompletion(),
+  linter(() => []),
```
microbit-matt-hillsdon, line 79: Do we need this?

src/editor/codemirror/linting/linter.worker.ts
```diff
@@ -0,0 +1,24 @@
...
+ctx.onmessage = (event: MessageEvent) => {
+  const errors = TPyParser.findAllErrors(event.data);
+  ctx.postMessage(
+    errors.map((error: any) => ({
+      offset: error.offset,
+      line: error.line,
+      msg: error.msg,
+    }))
+  );
+};
```
microbit-matt-hillsdon, line 24: This can go from this branch.

src/editor/codemirror/linting/tigerpython-parser.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * (c) 2021, Micro:bit Educational Foundation and contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+declare module "tigerpython-parser";
```
microbit-matt-hillsdon, line 7: This can go.

src/editor/codemirror/config.ts
microbit-matt-hillsdon: We need the linter as the language server client sets the diagnostics directly. Maybe we can reconsider but enough for now.

public/workers/pyright-a8ec35e78b1967df3217.worker.js.LICENSE.txt
```diff
@@ -0,0 +1,8 @@
+/*!
+ * The buffer module from node.js, for the browser.
+ *
+ * @author   Feross Aboukhadijeh <https://feross.org>
+ * @license  MIT
+ */
+
+/*! ieee754. BSD-3-Clause License. Feross Aboukhadijeh <https://feross.org/opensource> */
```
microbit-matt-hillsdon, line 8: Hah, this is clearly inadequate. It's just what the webpack build splits out. We'd be better with some explicit credit for Pyright.

I'd also like to look at publishing a package for it rather than repeatedly splatting 1.3MB of JS in this repo. That'd be a good chance to improve this.

src/editor/codemirror/linting/worker-transport.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.worker.removeEventListener("message", this.listener);
+  }
+
+  async sendData(
+    data: JSONRPCRequestData,
+    timeout?: number | null
+  ): Promise<any> {
+    const promise = this.transportRequestManager.addRequest(data, null);
+    this.worker.postMessage((data as IJSONRPCData).request);
+    return promise;
```
microbit-matt-hillsdon, line 33: I'd like to convince myself this is correct. There's lots of code in the language server client that doesn't await promises and I don't understand why.

src/mocks/worker.js
```diff
@@ -0,0 +1,5 @@
+// WebWorker stub for jest.
+
+class MockWorker {}
+
+module.exports = MockWorker;
```
microbit-matt-hillsdon, line 5: Not needed, I think.

tsconfig.json
```diff
@@ -1,11 +1,7 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["dom", "dom.iterable", "esnext"],
```
microbit-matt-hillsdon, line 4: Not really indented but if that's what prettier wants it's inevitable at some point.

src/editor/codemirror/linting/worker-transport.ts
microbit-matt-hillsdon: Investigated this a bit and following the flow of the code by hand I don't see how the rpc library doesn't leak a promise for every notification it sends. Needs proper debugging.

microbit-matt-hillsdon: It does leak. The design requires each transport to settle the promises for notifications, which isn't done in the postMessage based ones. I might be missing a reason why but as notifications are a requests without a response I think it's probably a bug.

Raised https://github.com/open-rpc/client-js/issues/294

jaustin:
Having a quick play with this alongside #214, there are definitely some cases that I think the TigerPython errors make much more clear

This error:
image

Compared to this one:
image

@microbit-matt-hillsdon - I know we've discussed it generally, but thought I'd add here: is it possible to get down to the nice specificity of TigerPython with Pyright?

src/editor/codemirror/linting/worker-transport.ts
microbit-matt-hillsdon: This is resolved if we go with #251. It also looks to be fixed upstream but I've not upgraded. There's quite a win from the VS Code JSON-RPC package's type support so I think it's the right route for us.

microbit-matt-hillsdon:
Work will slowly move to #251, but for now this branch has more features and is a useful reference.
I'll move the comments too before closing this.

microbit-matt-hillsdon:
> @microbit-matt-hillsdon - I know we've discussed it generally, but thought I'd add here: is it possible to get down to the nice specificity of TigerPython with Pyright?

This particular case was a bug in the language server integration. Pyright now reports `Expected ":"` at the same point as Tiger Python. It does also have some knock-on errors that seem less than ideal, including that one about the expected expression. I suspect this could be a bug. I'll take a look soon.

public/workers/pyright-a8ec35e78b1967df3217.worker.js.LICENSE.txt
microbit-matt-hillsdon: Sorted on #251

src/editor/codemirror/linting/worker-transport.ts
microbit-matt-hillsdon: It worked out well.

microbit-matt-hillsdon:
> This particular case was a bug in the language server integration [1]. Pyright now reports `Expected ":"` at the same point as Tiger Python. It does also have some knock-on errors that seem less than ideal, including that one about the expected expression. I suspect this could be a bug. I'll take a look soon.

Raised #257 to track "soon" 🙂

microbit-matt-hillsdon:
Closing as we're going with #251.

PR #224, opened 2021-08-09
Show runtime errors in compact serial view

microbit-matt-hillsdon:
- Add serial utility features from V2 for Ctrl-D, Ctrl-C. 
- Add some minimal in-app help to explain what's going on.

We've stayed away from inline presentation of errors in the editor as runtime errors naturally get out of sync with the current code and we feel it'll be more confusing then helpful. Work is planned for lint errors that will be shown inline as they'll be up-to-date.

There's little by way of e2e testing here. I've made a start on #226 as we'll want to mock the serial interactions.
I really want to cover the line/file navigation.

### DIFF
--- a/src/editor/codemirror/CodeMirror.tsx
+++ b/src/editor/codemirror/CodeMirror.tsx
@@ -108,7 +108,7 @@ const CodeMirror = ({
   }, [options]);
 
   useEffect(() => {
-    // When identify of location changes then the user has navigated.
+    // When the identity of location changes then the user has navigated.
     if (location.line) {
       const view = viewRef.current!;
       let line;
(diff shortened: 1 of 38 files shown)


src/editor/codemirror/CodeMirror.tsx
```diff
@@ -104,6 +107,26 @@ const CodeMirror = ({
     });
   }, [options]);
 
+  useEffect(() => {
+    // When identify of location changes then the user has navigated.
```
microbit-matt-hillsdon, line 111:
```suggestion
    // When the identity of location changes then the user has navigated.
```

PR #226, opened 2021-08-10
Introduce a mock device for e2e testing.

microbit-matt-hillsdon:
Might be nice to avoid the unconditional import but not so bad while
it's small.


microbit-matt-hillsdon:
Rebased on merged work for #224.

microbit-matt-hillsdon:
Branch preview still uses the real device 😄  Merging as a small step in the right direction.

Issue #148, opened 2021-06-08
[CodeMirror] Incremental parse bug


microbit-matt-hillsdon:
Delete/replace the colon after the last if True in the while. Notice the structural highlighting is broken for the last if/else.

Edit the last if/else and it fixes itself.

Needs minimising and a CodeMirror bug reporting (once we've checked we're on the latest).

```
# Add your Python code here. E.g.
from microbit import *


while True:
    display.scroll('Hello, World!')
    display.show(Image.HEART)
    sleep(2000)
    if True:
        pass
    if True:
        pass

if True:
    pass
    
if True:
    pass
else:
    pass
```

microbit-matt-hillsdon:
CM bug is fixed (unclear by what change, but doesn't reproduce with the latest deps)..

Next step is to upgrade CM and test.

microbit-matt-hillsdon:
Closed by #232.

microbit-matt-hillsdon:
Reopening as we've had to revert the CodeMirror upgrade due to #243.

microbit-matt-hillsdon:
We re-upgraded. Closing.

PR #232, opened 2021-08-11
Upgrades.

microbit-matt-hillsdon:
CM motivated by #148 

Quite a bit of CM change so we should review with care in addition to checking the bug fix.

Probably sensible to review the back catalog of incremental parse bugs as they're quick to test.


microbit-matt-hillsdon:
Seems fixed but we should fix #233 and retest to be more confident.

microbit-matt-hillsdon:
Other than code structure issues this seems to be working well.

microbit-matt-hillsdon:
Rebased on main with fix for #233 so once that's deployed I'll retest.

PR #251, opened 2021-08-24
LS based errors - multi-file branch

microbit-matt-hillsdon:
Attempt to improve on the client used in #222.

Fewer features for now, but what we have is more considered rather than a quick integration.

We're using the VS Code JSON-RPC support which comes with a nice approach to typing messages sent across the connection.

We attempt to implement multi-file support by opening (in a LSP-sense) all the documents so their contents are controlled by us and kept in sync with any app changes. This needed more changes in our Pyright fork which is covered by 

Remaining tasks:
- [x] Rebase
- [x] Review diagnostics at the end of a line. I think we're losing these, e.g. missing colon.
    -  Worked around and started a discussion here: https://discuss.codemirror.net/t/diagnostics-for-the-range-from-end-of-a-line-to-the-start-of-the-next/3495
    - [x] That message also has 4 run-on errors. Is that a Pyright bug? It looks like it's trying to do sensible error recovery. Track separately: #257 
- [x] Adequate typings so we can ship this.
    - Depending on how painful this is, we could use to a feature flag for now.
- Clean up Pyright PR and merge into our browser branch. Document any limitations etc. as it's quite involved.
    - I'm going to just track this on its own PR. I've done the user-visible clean-up (debug logging etc.) for this PR.
- [x] Implement autocomplete in the new client
- [x] Credit/licensing
- Enable `python.analysis.completeFunctionParens` ? We don't currently have a way to configure this. It is off by default for our users. Probably best split out.
    - I misunderstood, this is a closed-source Pylance feature. We can probably do this ourselves but it should definitely be split out. There's "kind" information on the completion items so we should know whether they're callable. We'll need to dispatch an editor state update that also moves the cursor between the brackets.

Leaving hover (for now) as per discussion on #222.

Demo at https://review-python-editor-next.microbit.org/lsp-rethink/


microbit-matt-hillsdon:
Possible abstraction over worker_threads and Web Workers: https://github.com/developit/web-worker

src/editor/codemirror/language-server/diagnostics.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const diagnosticsMapping = (
+  document: Text,
+  diagnostics: LSP.Diagnostic[]
+): Diagnostic[] =>
+  diagnostics
+    .map(({ range, message, severity }): Diagnostic | undefined => {
+      let from = positionToOffset(document, range.start);
+      let to = positionToOffset(document, range.end);
+      console.log(message, range.start, range.end, from, to);
```
microbit-matt-hillsdon, line 26: Debug.

src/editor/codemirror/language-server/diagnostics.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      // Skip if we can't map to the current document.
+      if (from !== undefined && to !== undefined) {
+        // If `from` is the end of one line and `to` is the beginning of another
+        // then CM won't display the diagnostic inline. Shift the `to` to also
+        // be at the end of the line.
```
microbit-matt-hillsdon, line 32: Add a note that we can remove once the commit referenced here is merged: https://discuss.codemirror.net/t/diagnostics-for-the-range-from-end-of-a-line-to-the-start-of-the-next/3495

src/editor/codemirror/language-server/view.ts
```diff
@@ -0,0 +1,82 @@
...
+          view.state,
+          diagnosticsMapping(this.view.state.doc, initialDiagnostics)
+        )
+      );
+    }, 0);
+  }
+
+  update({ docChanged }: ViewUpdate) {
+    if (docChanged) {
+      // TODO: incremental updates!
```
microbit-matt-hillsdon, line 52: Raise an issue for this.

microbit-matt-hillsdon: #256

Issue #223, opened 2021-08-06
Hiding browser tab during flash can leave flash progress modal up


microbit-matt-hillsdon:
I did a Python editor flash after MakeCode this morning and opened a new tab while it ran. The flash completed successfully but then I suspect we disconnected due to tab visibility which left this mess:

![image](https://user-images.githubusercontent.com/44397098/128485556-781fc7bf-7ba5-4285-948e-28e4ceee289f.png)

```
Detected board ID 9904
instrument.ts:129 Device status CONNECTED
instrument.ts:129 Stopping serial before flash
instrument.ts:129 Finished listening for serial data
instrument.ts:129 Reconnecting before flash
instrument.ts:129 Detected board ID 9904
instrument.ts:129 Device status CONNECTED
instrument.ts:129 Begin reset
instrument.ts:129 Begin flashing
instrument.ts:129 Total pages: 128
instrument.ts:129 Changed pages: 78
instrument.ts:129 Full flash
instrument.ts:129 Error during disconnection:
InvalidStateError: An operation that changes the device state is in progress.
instrument.ts:129 Device status NOT_CONNECTED
instrument.ts:129 Disconnection complete
instrument.ts:129 Error during disconnection:
InvalidStateError: An operation that changes the device state is in progress.
instrument.ts:129 Device status NOT_CONNECTED
instrument.ts:129 Disconnection complete
instrument.ts:129 Error during disconnection:
InvalidStateError: An operation that changes the device state is in progress.
instrument.ts:129 Device status NOT_CONNECTED
instrument.ts:129 Disconnection complete
instrument.ts:129 Device status NOT_CONNECTED
instrument.ts:129 Disconnection complete
```

I had a similar report from Jonny via Zoom.

microbit-matt-hillsdon:
When fixing this we might also want to consider the idea of making flashing state more available to the UI (e.g. we've discussed not saying "Running..." in the serial bar). Right now it's only the modal that's aware of what's going on. There are other incidental interactions due to the serial reset event.

microbit-matt-hillsdon:
Fairly convinced we just want to remove or disable the visibility listener during flash and that'll fix it.

But... there's a finally block that should prevent the dialog staying up regardless of errors. Why isn't that working?

microbit-matt-hillsdon:
As noted on #252, this didn't turn out to be easy to reproduce but does seem to be a logical flaw. I've added a fix and more logging in case we need to revisit in future.

PR #252, opened 2021-08-25
Avoid disconnect during flash.

microbit-matt-hillsdon:
Fixes logical bug that I can no longer reproduce in practice.

See https://github.com/microbit-foundation/python-editor-next/issues/223

### DIFF
--- a/src/device/device.ts
+++ b/src/device/device.ts
@@ -203,6 +203,8 @@ export class MicrobitWebUSBConnection
     this.emit(EVENT_SERIAL_DATA, data);
   };
 
+  private flashing: boolean = false;
+  private disconnectAfterFlash: boolean = false;
   private visibilityReconnect: boolean = false;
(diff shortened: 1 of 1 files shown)


src/device/device.ts
```diff
@@ -203,21 +203,30 @@ export class MicrobitWebUSBConnection
...
       if (
         this.visibilityReconnect &&
         this.status !== ConnectionStatus.CONNECTED
       ) {
+        this.disconnectAfterFlash = false;
         this.visibilityReconnect = false;
-        this.connect();
+        if (!this.flashing) {
+          this.connect();
+        }
```
microbit-matt-hillsdon, line 219: I think the flashing guard is redundant as we'll be connected if we're flashing.

microbit-matt-hillsdon: Though in some ways it makes it easier to reason about so I might leave it.

PR #255, opened 2021-09-06
Distinguish full flash in progress dialog.

### LINKED ISSUE #254: Flashing: explain to user that first flash takes a longer time
## Description

In the current live editor, we explain to the user that the first flash takes a longer time than subsequent flashes.

This is reassuring, given that the first flash can take a long time for slower machines
Screenshot2021_08_26_174957

## Todo

- [ ] Choose whether to lift message as is (I'm happy with that)
- [ ] Implement

microbit-matt-hillsdon:
Behaviour should match V2.

Closes #254.

### DIFF
--- a/src/project/project-actions.tsx
+++ b/src/project/project-actions.tsx
@@ -290,7 +290,7 @@ export class ProjectActions {
         id: "flashing-micropython",
       });
       const firstFlashNotice = (
-        <Text pb={4} fontSize="lg">
+        <Text fontSize="lg">
           <FormattedMessage id="flashing-full-flash-detail" />
         </Text>
       );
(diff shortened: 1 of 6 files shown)


src/project/project-actions.tsx
```diff
@@ -285,9 +285,19 @@ export class ProjectActions {
     }
 
     try {
-      const progress = (value: number | undefined) => {
+      const flashingCode = this.intl.formatMessage({ id: "flashing-code" });
+      const flashingMicroPython = this.intl.formatMessage({
+        id: "flashing-micropython",
+      });
+      const firstFlashNotice = (
+        <Text pb={4} fontSize="lg">
```
microbit-matt-hillsdon, line 293: Let's move the pb into the layout of the progress dialog (as mt) as it's necessary separation for anything we put above it.

microbit-matt-hillsdon: 6e5da6bf20e944eaeba9274bbdeeac4ad361a05b

PR #264, opened 2021-09-14
Upgrades.

microbit-matt-hillsdon:
Mainly to get @codemirror/view without a debug console.log message
that's quite distracting.


microbit-matt-hillsdon:
Concerned there might be an indent regression. I think this needs to wait until I catch up with the indent situation.

Issue #263, opened 2021-09-14
[API] Prototype API documentation in sidebar


microbit-matt-hillsdon:
An initial prototype of API documentation in the sidebar that integrates with the editor.

This is early, exploratory work.

Aim to explore:
- Where we get the data from (stubs?, via Pyright?, precompute it for quick rendering?)
- Whether the stubs contain enough information to do a good job. Is manual curation required?
- What documentation improvements are necessary or desirable
- How we display it
- How we interact with the editor (e.g. drag and drop, other)
- Maybe, whether there's any interaction the other way, e.g. focus/goto docs as an alternative to docs on hover

microbit-matt-hillsdon:
Status comment (I'll edit this as we go):

-  [Branch with API documentation document-style](http://review-python-editor-next.microbit.org/apidocs)
-  [Branch with API documentation as an accordion](http://review-python-editor-next.microbit.org/apidocs-accordian) (yes, I misspelled the branch 😞 )

Both pretty rough! We could focus on design / layout questions, documentation raising quality, drag-and-drop, progressive disclosure of doc string detail... 

If we're generally happy with the direction, and don't seem likely to switch to manually curated docs for this API section, then there's work to be done to tidy up the Pyright side of this too.

### Which modules?

I’ve left out modules for now that aren’t in our readthedocs docs: antigravity, builtins, love, this, errno, array, collections, gc, struct, sys.

From discussions with Giles it seems like we might want some docs for the builtins. The stubs currently don't have docs but they exist somewhere...

antigravity and love should probably remain undocumented easter eggs (you can autocomplete to them). The others are probably fine to add but perhaps we'd want to make some modules less prominent as we add more? Though I think a basic/advanced split is harder to justify.

### Docstrings

I've truncated them at the moment but that's not really helpful. They're not suitable as-is.

Many don't contain a short summary that we could extract.

There is no systematic approach to parameter documentation.

They contain a lot of restructured text artifacts which we don't yet attempt to format or remove.

It'd be good to discuss what to do here.

### Types

I've shown type information. I think it's useful in general but I'm definitely not the target audience. The very few places that use generics look arcane (e.g. random.choose).

### More minor

- Dunder methods / operations (e.g. adding Images) aren't covered.
- Various issues raised against the stubs.

jaustin:
@microbit-matt-hillsdon this looks excellent, thanks! I'm really surprised by how good it looks and how *not* overwhelming it is even in this early form. Very promising.

[...]

PR #269, opened 2021-09-23
Draft work exposing API docs in the sidebar

microbit-matt-hillsdon:
Additional, more beginner friendly documentation is also planned.

This is a draft. Discussion on #263


src/apidocs/ApiDocsArea.tsx
```diff
@@ -0,0 +1,230 @@
...
+    <>
+      <Accordion allowToggle wordBreak="break-word">
+        {sortBy(Object.values(docs), (m) => m.fullName).map((module) => (
+          <AccordionItem key={module.id}>
+            <AccordionButton
+              fontSize="xl"
+              _expanded={{ fontWeight: "semibold" }}
+            >
+              <Box flex="1" textAlign="left" mr={3}>
+                {module.fullName}
```
microbit-matt-hillsdon, line 57: This should probably be a heading of some sort. We should review everything beneath it for semantic html use.

src/apidocs/ApiDocsArea.tsx
```diff
@@ -0,0 +1,230 @@
...
+        </Box>
+      )}
+    </Box>
+  );
+};
+
+const groupHeading = (kind: string, childKind: string): string => {
+  switch (childKind) {
+    case "variable":
+      return "Fields";
```
microbit-matt-hillsdon, line 160: Here and elsewhere this needs strings with react-intl.

src/apidocs/ApiDocsArea.tsx
```diff
@@ -0,0 +1,230 @@
...
+      return "Classes";
+    case "function":
+      return kind === "class" ? "Methods" : "Functions";
+    default: {
+      throw new Error("Unexpected");
+    }
+  }
+};
+
+const formatName = (kind: string, fullName: string): string => {
```
microbit-matt-hillsdon, line 171: Everything has a name field now so perhaps we can avoid removing the last part of the full name.

src/apidocs/ApiDocsArea.tsx
```diff
@@ -0,0 +1,230 @@
...
+  // Add zero width spaces to allow breaking
+  return kind === "module"
+    ? fullName.replaceAll(/\./g, "\u200b.\u200b")
+    : fullName.split(".").slice(-1)[0];
+};
+
+const nameSuffix = (kind: string, type: string | undefined): string => {
+  if (kind === "function") {
+    return (type || "()").replace(/ -> None$/, "");
```
microbit-matt-hillsdon, line 180: This feels a bit arbitrary. We should resolve the discussion around how helpful types are here.

Another type that feels like clutter is the "self" parameter for instance methods. It wouldn't be shown in readthedocs.

src/apidocs/ApiDocsArea.tsx
```diff
@@ -0,0 +1,230 @@
...
+      array = [];
+      result.set(k, array);
+    }
+    array.push(v);
+  }
+  return result;
+}
+
+const BaseClasses = ({ value }: { value: ApiDocsBaseClass[] }) => {
+  const prefix = value.length === 1 ? "base class " : "base classes: ";
```
microbit-matt-hillsdon, line 211: Translation

src/apidocs/ApiDocsArea.tsx
```diff
@@ -0,0 +1,230 @@
...
+        <a key={bc.fullName} href={"#" + bc.fullName}>
+          {bc.name}
+        </a>
+      ))}
+    </Text>
+  );
+};
+
+const DocString = ({ value }: { value: string }) => (
+  <Text fontSize="sm" mt={2} noOfLines={2} fontWeight="normal">
```
microbit-matt-hillsdon, line 225: noOfLines is just a hack, we need to write good first lines of docstrings and then extract just those to show here

src/workbench/LeftPanelTabContent.tsx
```diff
@@ -23,7 +24,13 @@ const LeftPanelTabContent = ({
   return (
     <Flex height="100%" direction="column">
       {nav && <HStack justifyContent="flex-end">{nav}</HStack>}
-      <Box flex="1 0 auto" overflowY="auto" height={0}>
+      <Box
+        className="left-panel-content"
+        flex="1 0 auto"
+        overflowY={"overlay" as any}
```
microbit-matt-hillsdon, line 30: This was experimental and I don't think we mean to land it with this PR

src/documentation/AdvancedDocumentation.tsx
```diff
@@ -0,0 +1,34 @@
...
+      }
+    };
+    load();
+  }, [client]);
+  return (
+    <Box height="100%" p={2}>
+      {apidocs ? (
+        <ApiDocsAccordion docs={apidocs} />
+      ) : (
+        <Spinner label="Loading API documentation" alignSelf="center" />
```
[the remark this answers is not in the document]
microbit-matt-hillsdon: We might also can this initial JSON response for speed, but I still think we'll want to reload it when you add modules etc.

PR #274, opened 2021-09-30
Render markdown for the docs.

microbit-matt-hillsdon:
- Request it from the language server
- Render, minimally style etc.

The functionality that now uses this rendering is still behind feature flags.


src/language-server/client.ts
```diff
@@ -71,75 +71,83 @@ export class LanguageServerClient extends EventEmitter {
     return this.diagnostics.get(uri) ?? [];
   }
 
+  private initializePromise: Promise<void> | undefined;
+
+  /**
+   * Initialize or wait for in-progress initialization.
+   */
```
microbit-matt-hillsdon, line 78: Hmm, this init change wasn't intended. I just meant to request markdown.

The change is from the apidocs branch but not worth merging yet.

PR #275, opened 2021-09-30
Use markdown to render the apidocs

microbit-matt-hillsdon:
It's already in use for the signature and autocomplete help.

I'll target the apidocs branch with this change.

### DIFF
--- a/src/apidocs/ApiDocsArea.tsx
+++ b/src/apidocs/ApiDocsArea.tsx
@@ -18,8 +18,6 @@ import { pullModulesToTop } from "./apidocs-util";
 const ApiDocsArea = () => {
   const client = useLanguageServerClient();
   const [apidocs, setApiDocs] = useState<ApiDocsResponse | undefined>();
-  const [openId, setOpenId] = useState<string | undefined>(undefined);
-  console.log("Rendering top-level", openId);
   useEffect(() => {
     const load = async () => {
       if (client) {
@@ -34,7 +32,7 @@ const ApiDocsArea = () => {
   return (
     <Box height="100%" p={3} pt={4}>
       {apidocs ? (
-        <ModuleDocs docs={apidocs} openId={openId} setOpenId={setOpenId} />
+        <ModuleDocs docs={apidocs} />
       ) : (
         <Spinner label="Loading API documentation" alignSelf="center" />
       )}
@@ -43,12 +41,10 @@ const ApiDocsArea = () => {
 };
 
 interface ModuleDocsProps {
-  openId: string | undefined;
-  setOpenId: (id: string | undefined) => void;
   docs: ApiDocsResponse;
 }
 
-const ModuleDocs = ({ openId, setOpenId, docs }: ModuleDocsProps) => {
+const ModuleDocs = ({ docs }: ModuleDocsProps) => {
   return (
     <>
       {sortBy(Object.values(docs), (m) => m.fullName).map((module) => (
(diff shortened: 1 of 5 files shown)


src/apidocs/ApiDocsArea.tsx
```diff
@@ -224,4 +221,18 @@ const BaseClasses = ({ value }: { value: ApiDocsBaseClass[] }) => {
   );
 };
 
+const DocString = ({ value }: { value: string }) => {
+  const html = renderMarkdown(value);
+  console.log(value);
```
microbit-matt-hillsdon, line 226: Debug.

Something odd is going on. Most markdown is rendering OK, but the bold text is being returned with escaped asterisks. Perhaps this needs a change on the Pyright end. I can't figure out why this would happen though.

It's fine for the same documentation for signature help.

src/editor/codemirror/language-server/documentation.ts
```diff
@@ -30,3 +26,17 @@ export const renderDocumentation = (
...
+
+export const renderMarkdown = (markdown: string): SanitisedHtml => {
+  return {
+    __html: DOMPurify.sanitize(
+      render(markdown, {
+        gfm: true,
+      })
+    ),
+  };
+};
```
microbit-matt-hillsdon, line 42: Now we've exposed this directly it should have coverage to check for sanitisation.

src/apidocs/ApiDocsArea.tsx
```diff
@@ -224,4 +221,18 @@ const BaseClasses = ({ value }: { value: ApiDocsBaseClass[] }) => {
...
+  return (
+    <Text
```
microbit-matt-hillsdon, line 228: There's an issue with newlines that are \n in the reST. See the Image docs.

src/apidocs/ApiDocsArea.tsx
microbit-matt-hillsdon: Maybe I'm double-converting it to markdown on the sever side?

microbit-matt-hillsdon: Nope, this is just something that isn't supported in Pyright's doc string conversion.

Screenshot 2021-09-30 at 17 31 43

microbit-matt-hillsdon: I've hacked this for bold which we have a lot of, but it's clearly a hack as it has to work with escaped asterisks.

To fix this properly we'll need to spend some time with Pyright's docStringConversion code which will take a while to get into.

I've been giving some thought as to whether I should abandon the reSt -> markdown (in Pyright) -> HTML approach, but I think it's probably worth trying to fix these bugs as they'll lead to a better VS Code experience too. Even if we just end up tailoring the content somewhat.

I'll raise issues once this code ends up on main.

src/apidocs/ApiDocsArea.tsx
microbit-matt-hillsdon: I'll track this separately too, it's already broken when it comes back from Pyright.

src/apidocs/ApiDocsArea.tsx
```diff
@@ -224,4 +232,48 @@ const BaseClasses = ({ value }: { value: ApiDocsBaseClass[] }) => {
...
+              onClick={() => setOpen(!isOpen)}
+              rightIcon={<ExpandCollapseIcon open={isOpen} />}
+              _hover={{
+                textDecoration: "none",
+              }}
+              p={1}
+              pt={1.5}
+              pb={1.5}
+            >
+              {/* TODO: better aria-label with context */}
```
microbit-matt-hillsdon, line 269: TODO

PR #276, opened 2021-10-01
DnD support.

microbit-matt-hillsdon:
This might be of more value starting from other forms of documentation but I wanted to be able to demo this alongside apidocs.

There's no drop-cue but there's enough information in the state to draw one.


src/editor/codemirror/language-server/signatureHelp.ts
```diff
@@ -67,9 +65,15 @@ export class SignatureHelpView
       this.triggerSignatureHelpRequest();
     } else if (docChanged) {
       const last = transactions[transactions.length - 1];
-      if (last.isUserEvent("input.type")) {
+      if (
+        last.isUserEvent("input.type") ||
+        last.isUserEvent("input.drop.apidocs")
```
microbit-matt-hillsdon, line 70: Something ought to explain this is used to communicate triggering signature help.

microbit-matt-hillsdon:
Lots of this has made it back to main in a different form, but leaving open for the CM DND state management as we do plan to revisit dnd on #335

microbit-matt-hillsdon:
Closing this as I'm going to open a new draft PR, but keeping the branch around for now. We might want the decoration/state prep that's here.

PR #282, opened 2021-10-13
Style the autocomplete to be less visually noisy.

microbit-matt-hillsdon:


src/editor/codemirror/themeExtensions.ts
```diff
@@ -39,6 +39,28 @@ export const themeExtensions = (fontSizePt: number) => {
...
+    },
+    ".cm-tooltip.cm-completionInfo": {
+      width: "15rem",
+      height: "6rem",
+      top: "0 !important",
+      overflowY: "auto",
+    },
+    ".cm-tooltip-autocomplete > ul": {
+      height: "6rem",
+    },
```
microbit-matt-hillsdon, line 63: This whole block could do with an explanatory comment.

The DOM structure is weird here with tooltips nested inside each other which makes the borders a faff.

microbit-matt-hillsdon: Folks have commented that it should be taller/wider so I'll play around with that.

PR #294, opened 2021-10-18
Documentation links for autocomplete and signature help

microbit-matt-hillsdon:
The scrolling is awkward and doesn't play nice with the fixed accordion
header. We may well be changing that as part of design work.

TODO:
- [ ] Focus Advanced tab.
- [ ] Scrolling issue
    - Wait for / block on design changes that might avoid scrolling entirely?
- [ ] Less messy communication channel? Not clear how best to glue CM and React state together here.
- [ ] Hide when advanced docs not enabled (use flag).
- [ ] Translation in CM code (just "More info" or similar for now).

### DIFF
--- a/src/documentation/ApiDocsAccordion.tsx
+++ b/src/documentation/ApiDocsAccordion.tsx
@@ -4,10 +4,11 @@ import {
   AccordionIcon,
   AccordionItem,
   AccordionPanel,
+  ExpandedIndex,
 } from "@chakra-ui/accordion";
 import { Box, Text } from "@chakra-ui/layout";
 import sortBy from "lodash.sortby";
-import React from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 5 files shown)


src/documentation/ApiDocsAccordion.tsx
```diff
@@ -17,16 +18,47 @@ interface ApiDocsAccordionProps {
...
+      document.removeEventListener("cm/openDocs", listener);
+    };
+  }, [modules]);
+  useEffect(() => {
+    if (scrollTarget.current) {
+      const elt = document.getElementById(scrollTarget.current);
+      console.log("Scrolling", elt);
+      scrollTarget.current = undefined;
+      elt?.scrollIntoView();
+    }
```
microbit-matt-hillsdon, line 48: This doesn't scroll unless the index changed, which isn't what we want.

It also feels like I'm missing the simple way to do this! We need to know the render is done in the case where the index does change, hence the second effect but it's not very lovely.

Issue #277, opened 2021-10-10
pasting multi-line code into serial console results in wrong indentation


pdg137:
## Bug Description

If you paste properly indented code into the serial console, its auto-indentation hopelessly messes up the paste. Also, the command won't run after pressing "Enter". So it's hard to present examples for students to paste into their console.

## How To Reproduce

1. Copy this snippet to your clipboard:
    ```python
    if True:
      print("true")
    else:
      print("false")
2. Go to the serial console.
3. Paste with Ctrl+Shift+V.
4. Press enter.

## Expected behavior

It should look like this:

![image](https://user-images.githubusercontent.com/466760/136712668-ea74860e-ec8c-446d-9bc5-c057913a80cf.png)

(All automatic editing features should be disabled during a paste.)

## Actual behavior

![image](https://user-images.githubusercontent.com/466760/136712686-c5c5a0b6-9225-476f-80a4-ee653b8ab6d9.png)

## Environment

Latest Windows 10 / Brave browser.

pdg137:
Hmm, looks like multi-line paste doesn't do quite what I wanted:

![image](https://user-images.githubusercontent.com/466760/137563713-db3a4da3-46af-412b-be58-fced046247d1.png)

Is it actually for entering a single-line command across multiple lines?

![image](https://user-images.githubusercontent.com/466760/137564209-bc43864e-e7fa-4196-8ec2-e31d7388e0ea.png)

PR #295, opened 2021-10-19
Bracket multi-line pastes with Ctrl-E/Ctrl-D.

microbit-matt-hillsdon:
This is a proof of concept. If we want the feature we should raise an
issue against xterm.js asking for a way to intercept paste or disable
their paste handler as the current approach is very vulnerable to
internal changes in xterm.js.

I've also added a notification explaining what's happened.

See #277

### DIFF
--- a/src/serial/serial-hooks.tsx
+++ b/src/serial/serial-hooks.tsx
@@ -121,7 +121,8 @@ const useNewTerminal = (): Terminal => {
     };
 
     // Fragile interception of paste events.
-    // We should see if we can get API for this in xterm.js.
+    // Raised https://github.com/xtermjs/xterm.js/issues/3516
+    // for API in xterm.js
     const coreTerminal = (terminal as any)._core;
     const initGlobal = coreTerminal._initGlobal.bind(coreTerminal);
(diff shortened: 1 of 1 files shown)


src/common/use-action-feedback.tsx
```diff
@@ -81,6 +81,29 @@ export class ActionFeedback {
     });
   }
 
+  /**
+   * For when an action succeeds.
```
microbit-matt-hillsdon, line 85: Copy/paste comment error.

src/serial/serial-hooks.tsx
```diff
@@ -117,13 +119,44 @@ const useNewTerminal = (): Terminal => {
       terminalOpen(parent);
       resizeObserver.observe(parent);
     };
+
+    // Fragile interception of paste events.
+    // We should see if we can get API for this in xterm.js.
+    const coreTerminal = (terminal as any)._core;
```
microbit-matt-hillsdon, line 125:
```suggestion
    // Fragile interception of paste events.
    // Raised https://github.com/xtermjs/xterm.js/issues/3516
    // for API in xterm.js
    const coreTerminal = (terminal as any)._core;
```

PR #297, opened 2021-10-25
Advanced API docs early design and toolkit mock-ups

microbit-matt-hillsdon:
This takes something like the work-in-progress designs and applies it to the Advanced section.

We use the same design to mock-up the micro:bit and Python toolkits. The content in these sections is a partial placeholder. All code here is behind flags except that code that reacts to the flags.

TODO:
- [x] Scroll to top when the content of the sidebar tabs changes. To do this we need a ref to the scroll container, so we might want to push that down to the same level that's currently got the router state, then a simple effect can link the two? It's currently shared further up the tree, but that might be more confusing than helpful. See also the scroll handling for the code pop-up.
- [x] Initial implementation of "Insert code" for the micro:bit/Python toolkits. Some useful code on the apidocs-dnd branch.
- [x] Make it clearer that the micro:bit and Python sections are mock data, and ensure they're behind flags, so we can consider merging back behind a flag.
- [x] e2e tests, particularly for the autocomplete and signature help cross-links
- [x] Fix links to class fields/attributes (need a tested function to resolve a dotted name in the API docs as it's non-trivial).

Moved other todos to #263 and will split out into issues soon.


src/documentation/ToolkitDocumentation/TopicItem.tsx
```diff
@@ -0,0 +1,163 @@
...
+              value={code}
+              codeRef={codeRef}
+            />
+          )}
+        </Box>
+        <HStack spacing={3}>
+          <Button
+            fontWeight="normal"
+            color="white"
+            borderColor="rgb(141, 141, 143)"
```
microbit-matt-hillsdon, line 73: These colours etc. are hardcoded for now - I'll fix this when we have the actual design and I can better appreciate how they fit in with the theme.

src/documentation/ToolkitDocumentation/TopicItem.tsx
```diff
@@ -0,0 +1,163 @@
...
+            bgColor="rgb(141, 141, 143)"
+            borderTopRadius="0"
+            borderBottomRadius="xl"
+            variant="ghost"
+            size="sm"
+            onClick={() => actions?.insertCode(item.code)}
+          >
+            Insert code
```
microbit-matt-hillsdon, line 81: Similarly, I'm leaving translation of this for now until the design is finalised.

PR #299, opened 2021-10-30
Generalise import handling to work with Python code

microbit-matt-hillsdon:
Generalise import handling to work with Python code specifying the imports rather than a restrictive JSON model of the imports. This way we're closer to being able to add arbitrary Python snippets to existing code respecting their imports.


microbit-matt-hillsdon:
PR is just into another branch and all this code is unused so far. Merging.

PR #303, opened 2021-11-02
Update selection when inserting code.

microbit-matt-hillsdon:


src/editor/active-editor-hooks.tsx
```diff
@@ -51,8 +51,15 @@ export class ActiveEditorActions {
    */
   insertCode = (code: string): void => {
     const state = this.view.state;
+    const changes = calculateChanges(state, code);
+    const changeSet = state.changes(changes);
+    const lastChange = changes[changes.length - 1];
+    const updatedSelection =
+      changeSet.mapPos(lastChange.from) + lastChange.insert.trim().length;
```
microbit-matt-hillsdon, line 58: We should just remove trailing whitespace here.

PR #308, opened 2021-11-04
Add an introduction.

microbit-matt-hillsdon:
This matches the content. Design up for discussion.


microbit-matt-hillsdon:
There's a bit too much space below this. Pushed a tweak.

PR #312, opened 2021-11-07
Font tweaks for consistency.

microbit-matt-hillsdon:
Use the same larger font size for the detail view in Reference as we do
in the other toolkits.

### DIFF
--- a/src/documentation/DocString.tsx
+++ b/src/documentation/DocString.tsx
@@ -3,24 +3,18 @@
  *
  * SPDX-License-Identifier: MIT
  */
-import { Box } from "@chakra-ui/layout";
+import { Box, BoxProps } from "@chakra-ui/layout";
 import React from "react";
 import { renderMarkdown } from "../editor/codemirror/language-server/documentation";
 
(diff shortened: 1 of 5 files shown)


src/documentation/DocString.tsx
```diff
@@ -3,24 +3,18 @@
...
 }
 
-const DocString = React.memo(({ value }: DocStringProps) => {
+const DocString = React.memo(({ value, ...props }: DocStringProps) => {
   const html = renderMarkdown(value);
   return (
-    <Box
-      className="docs-markdown"
-      fontSize="sm"
-      mt={2}
```
microbit-matt-hillsdon, line 20: Had removing these affected any other use of DocString?

microbit-matt-hillsdon: They for sure don't belong there now it's been extracted but still needs review.

microbit-matt-hillsdon: Addressed in 9001d9e14b86b3324e50b4d7dc8e8b1c15273ffc

PR #316, opened 2021-11-15
Populate toolkits with (work-in-progress/incomplete) CMS content

### LINKED ISSUE #315: [Toolkit] Pull mock data from CMS replacing JSON mock data
We have an initial model of the toolkit data in the micro:bit CMS. Some content modelling tweaks might be required but it has enough structure and example content to replace the toolkit-mock JSON file and allow us to fill it out while seeing the design.

GROQ query that pulls all the data:

```
*[_type == "toolkit"]{id, name, description, contents[]->{name, subtitle, introduction, contents[]->{name, content, alternatives, detailContent}}}
```

[The same as a URL](https://ajwvhvgo.api.sanity.io/v1/data/query/apps?query=*%5B_type%20%3D%3D%20%22toolkit%22%5D%7Bid%2C%20name%2C%20description%2C%20contents%5B%5D-%3E%7Bname%2C%20subtitle%2C%20introduction%2C%20contents%5B%5D-%3E%7Bname%2C%20content%2C%20alternatives%2C%20detailContent%7D%7D%7D).

We could split up the fetching, but in the long run we might move away from directly pulling this directly from the CMS and we're interested in offline mode, so a bulk load probably makes sense if the size is reasonable.

We should aim to have a good basis to evolve the content model without breaking changes before merging work that depends on it.

microbit-matt-hillsdon:
Closes #315. We'll raise more issues for the details.

TODO on this PR:
- [x] e2e test failure
- [x] i18n
- [x] review of error / loading states
    - I think it's probably time to formalise this via storybook, I'm just sad about the fact it'll be painful to upgrade in future. I'll do that as a separate PR.
- [x] final review of CMS schema
    - Remaining issues relate to linking and I think can be tackled additively on a separate issue.

To raise issues for:
- [x] linking / navigation between toolkit entries
    - Raised https://github.com/microbit-foundation/python-editor-next/issues/318


src/documentation/ToolkitDocumentation/TopicItem.tsx
```diff
@@ -42,96 +59,183 @@ const TopicItem = ({
...
+          ...state,
+          // Hmm, we need to know the tab/toolkit (we should name them the same).
+          // We also need to switch to router-based navigation for the other toolkits.
+        });
+      }}
+    >
+      {props.children}
+    </Link>
+  );
+};
```
microbit-matt-hillsdon, line 180: This is all temporary and will be revisited on its own issue.

src/documentation/ToolkitDocumentation/model.ts
```diff
@@ -5,44 +5,55 @@
...
-  _type: "text";
-  detail?: boolean;
-  value: string;
+  _type: "block";
+}
+
+export interface ToolkitImage {
+  _type: "simpleImage";
+  // More!
+}
```
microbit-matt-hillsdon, line 39: In practice we're not using these types. Let's revisit the typing here and aim to be more consistent - it might be that a marker for portable text is all that's worth doing and we can keep it untyped.

microbit-matt-hillsdon: Or we use them in the serializers to type `node`.

microbit-matt-hillsdon: Rationalised in d5c801b85315bff8b6e8e563d29c3bd700da3723

src/documentation/ToolkitDocumentation/TopicItem.tsx
microbit-matt-hillsdon: https://github.com/microbit-foundation/python-editor-next/issues/318

PR #324, opened 2021-11-16
Data logging - preview feature in alpha editor

microbit-matt-hillsdon:


src/editor/codemirror/language-server/documentation.ts
```diff
@@ -52,8 +52,9 @@ const fixupMarkdown = (input: string): string => {
   // Messy because it's after escaping. Fragile because it's regex.
   // Let's see if we can upstream or align the docs with supported syntax.
   return input
-    .replace(/^\\\n/, "")
```
microbit-matt-hillsdon, line 55: I don't think I meant to do this.

src/editor/codemirror/language-server/documentation.ts
```diff
@@ -52,8 +52,9 @@ const fixupMarkdown = (input: string): string => {
...
     .replace(/`([\w² \n]+?) ?<(.*?)>`\\_/gs, "[$1]($2)")
+    .replace(/\\\*args/, "*args")
+    .replace(/\\\*kwargs/, "*kwargs")
```
microbit-matt-hillsdon, line 57: Need to check this was required.

This whole thing is pretty fragile, but I think less costly than improving the rEST -> Markdown code in Pyright. Worth revisiting if/when we use third-party docstrings.

microbit-matt-hillsdon: Required to avoid leading slash in the *args parameter.

src/editor/codemirror/language-server/documentation.ts
microbit-matt-hillsdon: Reverted.

microbit-matt-hillsdon:
I used the short hash in the about dialog, which looks OK:

image

microbit-matt-hillsdon:
This looks good from a code point of view. We'll be doing some internal testing with this branch and adding toolkit content for data logging.

microbit-matt-hillsdon:
Agreed with Jonny that we'd merge this today.

PR #337, opened 2021-11-26
Fix import line break issue

### LINKED ISSUE #336: [Toolkits] Insert code line break issue with blank editor
image

Code in the CMS is sensible, so this is a bug in our handling of the imports:

```python
from microbit import *
import music


display.scroll('score', delay=100, loop=True, wait=False) 
music.play(music.ODE)
```

Should be easy to repro in a unit test.

microbit-matt-hillsdon:
Closes #336

### DIFF
--- a/src/editor/active-editor-hooks.tsx
+++ b/src/editor/active-editor-hooks.tsx
@@ -50,18 +50,7 @@ export class ActiveEditorActions {
    * @param code The code with any required imports.
    */
   insertCode = (code: string): void => {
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
Changes make sense and and the tests pass, just need to try this one for real.

microbit-matt-hillsdon:
This works "for real" but the cursor position isn't at the end of the inserted text. Not sure if this broke that, but worth looking at before merging.

PR #338, opened 2021-11-26
Only filter out the microbit wildcard import.

microbit-matt-hillsdon:


src/documentation/ToolkitDocumentation/TopicItem.tsx
```diff
@@ -248,11 +252,20 @@ const CodeEmbed = ({
   const [hovering, setHovering] = useState(false);
   // Strip the imports.
```
microbit-matt-hillsdon, line 253: This isn't really needed anymore. The code actually removing imports is clear and overall it is more nuanced now.

microbit-matt-hillsdon, line 253:
```suggestion
```

microbit-matt-hillsdon:
This is good to merge after the rebuild. 

It might change after feedback as there's not been much opportunity to discuss, but it's a safer starting point and addresses user confusion about copy/pasting snippets that require e.g. music or radio.

PR #344, opened 2021-11-29
Align component names with current terminology.

microbit-matt-hillsdon:
Mostly Reference/Explore, but also "LeftPanel" -> "SideBar" as although
it's not user visible it's what we've been calling it in discussions for
some time.


src/documentation/ExploreArea.tsx
```diff
@@ -1,23 +1,19 @@
...
+import { Toolkit } from "./ToolkitDocumentation/model";
+import ToolkitSpinner from "./ToolkitDocumentation/ToolkitSpinner";
 
 type State =
   | { status: "ok"; toolkit: Toolkit }
   | { status: "error" }
   | { status: "loading" };
 
-interface ToolkitContainerProps {
-  toolkitId: string;
```
microbit-matt-hillsdon, line 17: This has been unused since we switched to a single toolkit.

PR #345, opened 2021-11-29
Always reset tabs when clicking on the tab header

### LINKED ISSUE #340: [Toolkits] Clicking tab should reset tab to top level
It's been proposed that clicking on a tab (e.g. Reference, Explore) should reset the tab to the top level.

This sounds OK in isolation, but I think I was ultimately expecting it to toggle the visibility of the sidebar, as in VS Code.

One to discuss.

microbit-matt-hillsdon:
As requested by Jonny / Giles.

Closes #340


microbit-matt-hillsdon:
Leaving open until I've checked we intend to reset for all clicks on tabs not just when the tab is already open.

PR #346, opened 2021-11-29
Remove the toolkit flag.

microbit-matt-hillsdon:
We still need to reinstate some kind of alpha notice.
I'll experiment with UI for that next.
I've left the feedback form etc. in case it's useful.


src/flags.ts
```diff
@@ -12,13 +12,13 @@
 
 import { stage } from "./environment";
 
-type Flag = "signatureHelp" | "toolkit";
+type Flag = "signatureHelp";
```
microbit-matt-hillsdon, line 15: We should really get rid of this one too!

microbit-matt-hillsdon: Not in this PR, but I'll raise it again for discussion.

microbit-matt-hillsdon:
Code is fine, just need to agree UX of feedback is good enough.

microbit-matt-hillsdon:
Rebased and resolved conflict with #350. Still blocked on discussion but hoping we can agree the feedback flow is good enough to put toolkits live in alpha.

microbit-matt-hillsdon:
OK, UI changes are approved, it's just the impact on e2e to fix now. Not sure why the localhost check isn't doing the trick.

src/editor/codemirror/language-server/documentation.ts
```diff
@@ -93,28 +92,26 @@ export const wrapWithDocumentationButton = (child: Element, id: string) => {
   wrapper.appendChild(child);
 
   // Nothing to link to without this flag.
```
microbit-matt-hillsdon, line 94: Flag is gone so comment should be too!

```suggestion
```

src/workbench/SideBar.tsx
```diff
@@ -143,7 +133,13 @@ const SideBarContents = ({ panes, ...props }: SideBarContentsProps) => {
       >
         <TabList>
           {/* Top margin aims to align with other logo. */}
-          <Box width="3.75rem" mt="1.2rem" ml="auto" mr="auto" mb="11.5vh">
+          <Box
+            width="3.75rem"
+            mt="1.2rem"
+            ml="auto"
+            mr="auto"
+            mb="max(11.5vh, 7.7rem)"
```
microbit-matt-hillsdon, line 141: This change prevents the curve of the first tab overlapping with the coloured area at the top of the toolkit. Will revisit when we get search designs.

microbit-matt-hillsdon:
Code change looks good and e2e issue is fixed. Just need CI to pass one last time and we're good to merge.

Issue #332, opened 2021-11-19
[Toolkits] Allow formatting in toolkit topic introduction


microbit-matt-hillsdon:
The toolkit topic introduction is currently plain text. We should allow all the formatting allowed in the other content areas. In particular we have one that uses an image and many that explain required imports which needs the code font and new paragraphs.

microbit-matt-hillsdon:
Seems worth doing this before the toolkit changes go live, as then I don't need to worry about breakage.

microbit-matt-hillsdon:
Switched to the same content type as the other sections. We might not use all its features but let's leave the option open for now.

I've added in obviously missing `code` formatting which is the only difference you can see in the app. I've not revisited the source content, which doubtless has a few other bits I couldn't previously represent. Reviewing the comments there is on my list for this week anyway so I'll deal with that separately.

Closing.

PR #351, opened 2021-11-29
The introduction is now always portable text.

microbit-matt-hillsdon:
Tidying after migration for https://github.com/microbit-foundation/python-editor-next/issues/332


microbit-matt-hillsdon:
This looks good locally and the code change seems fine. I'll review the PR branch in the morning.

PR #353, opened 2021-11-30
Fix unintended full lodash import.

microbit-matt-hillsdon:
Should remove ~70k from the bundle report.


microbit-matt-hillsdon:
Bundle size is down. Merging.

PR #354, opened 2021-11-30
Bigger click target for "back".

microbit-matt-hillsdon:


microbit-matt-hillsdon:
We'll doubtless revisit this again with the search designs but this is a big improvement.

PR #355, opened 2021-11-30
Remove unused prop.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
This was added when we tried using code font for module names in the Reference tab, but we backed away from it as it looked unnecessarily scary.

PR #360, opened 2021-12-07
Remove the signature help flag.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
Code change looks good, just needs to rebuild.

Issue #335, opened 2021-11-22
[Toolkits] DnD from Reference and Explore toolkits


microbit-matt-hillsdon:
Implement DnD from the Reference and Explore toolkits into the code editor.

- On hover drag handles on the code snippets in Explore and the signature in Reference (tablet to be considered later, might require more changes).
- "Ghost code" in the editor, similar to previous spike.
- Drop behaviour as per the apidocs-dnd branch for Reference drags. For Explore, just insert and move caret to the end.
- Show imports on drag? They're fixed regardless of where you drop could be clearer.

Desirable but later:
- Drop into the REPL. Does that affect showing imports on drag? Does it send Ctrl-C, then paste with paste mode.

For discussion after:
- Insert into REPL button when connected via WebUSB. Perhaps "Run" or "Try it" ?

microbit-matt-hillsdon:
For the ghost code, I think we can make edits to the document annotating the transaction with the `Transaction.addToHistory.of(false)` annotation. 

This is done the collab module which has a similar problem of edits from two sources. So long as we take responsibility for undoing our changes we'll be fine. We're playing on easy mode compared to collab as we know no other changes can happen during the drag operation.

microbit-robert:
Possible future improvement:

- When dragging and dropping a code snippet into an area that should be a code block (i.e., underneath `while True:` or something similar), this code snippet should be indented to proper level, instead of falling back to the base line (no indentation).

Issue #365, opened 2021-12-09
[Toolkits] Insert code can split lines


microbit-matt-hillsdon:
```python
from microbit import * #foo
```

The "Insert code" action in the sidebar is literally inserting the non-import code after the last import node, but that might not be the whole line.

In the scenario above it will split `#foo` onto its own line, which isn't desirable.

I might address this during work on #335 but raising here in case that doesn't happen.

microbit-matt-hillsdon:
This was resolved in the initial work on #335. Closing.

PR #366, opened 2021-12-09
Drag and drop of code snippets (initial PR, UI & behaviour behind flag)

microbit-matt-hillsdon:
First step towards #335 for the Explore toolkit.

Adds drag handles and a drag and drop experience between the Explore toolkit and the editor.

Drag handle UI is temporary.

UI and behaviour is behind a flag.

To tackle separately on #335:
- Styling for ghost text (non-trivial, requires adding decorations to the text, separate PR)
- Hide code pop-over during drag? Separate PR.
- Similar support for references? Separate PR.
- Discuss disabling diagnostics during drag. It's somewhat useful but very busy.

To verify behaviour is behind a flag you can use https://review-python-editor-next.microbit.org/dnd-2/?flag=none


src/editor/codemirror/edits.ts
```diff
@@ -89,29 +93,12 @@ export const calculateChanges = (state: EditorState, addition: string) => {
...
-    let insertionPoint = 0;
-    if (lastImport?.nextSibling) {
-      relation = Relation.Before;
-      insertionPoint = lastImport.nextSibling.from;
-    } else if (lastImport) {
-      relation = Relation.After;
-      insertionPoint = lastImport.to;
-    }
-
+    const line = state.doc.lineAt(pos!);
```
microbit-matt-hillsdon, line 97: We need to reinstate code that works when pos is undefined (as it is when clicking "Insert code"). I've dropped this as it wasn't line oriented and clearly needs to be (see #365). However, it's broken lots of tests, so either a quick fix or reinstate...

src/editor/codemirror/edits.ts
```diff
@@ -63,7 +63,11 @@ class AliasesNotSupportedError extends Error {}
  * @returns The necessary changes suitable for a CodeMirror transaction.
  * @throws AliasesNotSupportedError if the additional code contains alias imports.
  */
-export const calculateChanges = (state: EditorState, addition: string) => {
+export const calculateChanges = (
+  state: EditorState,
+  addition: string,
+  pos?: number
```
microbit-matt-hillsdon, line 69: We need tests for this new parameter when I've settled the rules for inserting.

src/editor/codemirror/dnd.ts
```diff
@@ -0,0 +1,77 @@
...
+
+// No access to the content via the event in dragover (as it may be cross-document)
+// so we need shared state with the drag so we can draw previews of the drop in
+// the document.
+let draggedCode: string | undefined;
+let invertedChanges: ChangeSet | undefined;
+
+export const setDraggedCode = (value: string | undefined) => {
+  draggedCode = value;
+};
```
microbit-matt-hillsdon, line 14: This kinda sucks, but I'm not sure it's worth the complication of avoiding the globals for this, particularly given it is part React and part CM. Leaving this open to ponder.

src/editor/codemirror/dnd.ts
```diff
@@ -0,0 +1,77 @@
...
+          if (pos !== null && pos !== lastPos) {
+            revertPreview(view);
+            lastPos = pos;
+            const transaction = calculateChanges(view.state, draggedCode, pos);
+            // Stash the inverse for when we next update the doc — we'll "undo" first.
+            invertedChanges = transaction.changes.invert(
+              transaction.startState.doc
+            );
+            // Skip selection updates
+            view.dispatch({
```
microbit-matt-hillsdon, line 41: Would it help to combine the revertPreview transaction with this one?

src/editor/codemirror/dnd.ts
```diff
@@ -0,0 +1,77 @@
...
+      drop(event, view) {
+        revertPreview(view);
+
+        event.preventDefault();
+        const pos = view.posAtCoords(event);
+        if (pos !== null) {
+          const pythonText = event.dataTransfer?.getData(
+            pythonSnippetMediaType
+          );
+          if (pythonText) {
```
microbit-matt-hillsdon, line 69: The order of this all seems wrong. Media type, then preventDefault, then position?

src/editor/codemirror/dnd.ts
microbit-matt-hillsdon: We could try this and see if it improves performance when throttling. I don't see the distinct states in the view so perhaps they're being collapsed by CM anyway.

src/editor/codemirror/dnd.ts
microbit-matt-hillsdon: Rearranged as suggested above.

src/editor/codemirror/dnd.ts
microbit-matt-hillsdon: I think this is necessary for correctness. We're reading a position, reverting the preview, then using that position with the new state after the revert. So we either map it or collapse the changes into one transaction.

microbit-matt-hillsdon: I've mapped it for now so the logic for computing the changes doesn't have to account for the other edit. Resolving.

microbit-matt-hillsdon:
End of document issue:

- The drop doesn't actually happen, the preview remains in place (and can't be undone)
- The drop effect indicator flickers

This will be when the line we're over isn't in the original document so we should be assuming it to be the last line.

src/editor/codemirror/edits.ts
microbit-matt-hillsdon: Added in 579417f0d30fb0d4c52b9f27e51d59b73132a9d7

package.json
```diff
@@ -59,9 +59,9 @@
     "@testing-library/user-event": "^12.8.3",
     "@types/file-saver": "^2.0.3",
     "@types/jest": "^26.0.24",
-    "pptr-testing-library": "^0.6.4",
+    "pptr-testing-library": "^0.7.0",
     "prettier": "2.3.2",
-    "puppeteer": "^8.0.0",
+    "puppeteer": "^13.0.0",
```
microbit-matt-hillsdon, line 64: Upgraded to get support for testing drag and drop.

src/e2e/app.ts
```diff
@@ -659,6 +673,40 @@ export class App {
...
+    const section = await (heading.getProperty("parentNode") as Promise<
+      ElementHandle<Element>
+    >);
+    const draggable = (await section.$("[draggable]"))!;
+    const lines = await document.$$("[data-testid='editor'] .cm-line");
+    const line = lines[targetLine - 1];
+    if (!line) {
+      throw new Error(`No line ${targetLine} found. Line must exist.`);
+    }
+    const findPoint = (e: Element) => {
```
microbit-matt-hillsdon, line 699: Rename to topLeft? Move comment to use?

src/editor/codemirror/dnd.ts
```diff
@@ -0,0 +1,126 @@
+import { ChangeSet, Transaction } from "@codemirror/state";
+import { EditorView } from "@codemirror/view";
+import { pythonSnippetMediaType } from "../../common/mediaTypes";
+import { calculateChanges } from "./edits";
+
+/**
+ * Information stashed last time we handled dragover.
+ * Cleared on drop of dragleave.
```
microbit-matt-hillsdon, line 8: Typo.

src/editor/codemirror/dnd.ts
```diff
@@ -0,0 +1,126 @@
...
+   * The last drag position.
+   */
+  line: number;
+  /**
+   * The inverse set of changes to the changes made for preview.
+   */
+  previewUndo: ChangeSet;
+}
+
+let lastDragPos: LastDragPos | undefined;
```
microbit-matt-hillsdon, line 21: Pretty sure at some point I'd moved this inside the dndSupport so it was per-editor. That seems more defensive to me. I think if there were multiple editors the events would happen in the right order for it not to be an issue, but an easy improvement anyway.

src/editor/codemirror/edits.ts
```diff
@@ -77,59 +73,93 @@ export const calculateChanges = (state: EditorState, addition: string) => {
...
-      relation = Relation.Before;
-      insertionPoint = lastImport.nextSibling.from;
-    } else if (lastImport) {
-      relation = Relation.After;
-      insertionPoint = lastImport.to;
+    let additionPrefix = "";
+    if (line !== undefined) {
+      // Tweak so the addition preview is under the mouse even if we added imports.
+      line -= importLines;
+      line = Math.max(1, line);
```
microbit-matt-hillsdon, line 98: Clearer as one line without `-=`.

src/editor/codemirror/edits.ts
```diff
@@ -77,59 +73,93 @@ export const calculateChanges = (state: EditorState, addition: string) => {
...
+      const extraLines = line - state.doc.lines;
+      if (extraLines > 0) {
+        additionInsertPoint = state.doc.length;
+        additionPrefix = "\n".repeat(extraLines);
+      } else {
+        additionInsertPoint = state.doc.line(line).from;
+      }
+    } else {
+      additionInsertPoint = skipWhitespaceLines(
```
microbit-matt-hillsdon, line 107: I can't really remember why we do this. Remove it and see what fails 😄

microbit-matt-hillsdon: This is covered by the "non-import content before code rather than after imports" case. I've added an explanatory comment.

microbit-robert (commented):
When dragging and dropping the same code snippet into the editor, it sometimes replaces the previously dropped changed. I can reproduce this, but not consistently. I think when this occurs, it also causes problems with the history (undo/redo). This can also result in the 'drop' failing to register completely. In my experience, this has typically occurred after moving to a different area within the 'Explore' docs after the initial error occurs.

microbit-matt-hillsdon:
> When dragging and dropping the same code snippet into the editor, it sometimes replaces the previously dropped changed. I can reproduce this, but not consistently. I think when this occurs, it also causes problems with the history (undo/redo). This can also result in the 'drop' failing to register completely. In my experience, this has typically occurred after moving to a different area within the 'Explore' docs after the initial error occurs.

Struggled to reproduce anything for a while but I have seen this:

```
dragstart
   dragover Line {from: 394, to: 427, number: 19, text: 'display.scroll(compass.heading())'}
   dragover Line {from: 428, to: 461, number: 20, text: 'display.scroll(compass.heading())'}
   dragleave (ignored)
   dragover Line {from: 462, to: 462, number: 21, text: ''}
   dragleave DOMRect {x: 549, y: 57, width: 716, height: 746.484375, top: 57, …} {x: 0, y: 0}
dragend
```

Unsure why the x, y are zero, but that, combined with the way we detect a "real" dragleave is one problem at least.

microbit-robert:
I've attached the logging outputs from the three issues I've seen: a drop not registering; a dropped snippet seemingly replacing a previously dropped snippet; breaking the drop feature.

[drop-not-registered.txt](https://github.com/microbit-foundation/python-editor-next/files/7726266/drop-not-registered.txt)
[breaking-drop-function.txt](https://github.com/microbit-foundation/python-editor-next/files/7726267/breaking-drop-function.txt)
[replacement-of-existing.txt](https://github.com/microbit-foundation/python-editor-next/files/7726268/replacement-of-existing.txt)

microbit-robert (approved):

PR #382, opened 2021-12-19
Avoid drag handle repaint due to height change.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
The editor itself still glitches — I'll look at that separately.

microbit-robert (approved):
LGTM

Issue #364, opened 2021-12-09
Initial content to Crowdin for translation


microbit-matt-hillsdon:
- [x] UI strings
- [x] Explore toolkit content

We'll probably want to make a lot of changes to the source Explore toolkit content, so we'll have a manually scripted sync with Crowdin rather than the CI-based process used in other Foundation projects.

microbit-matt-hillsdon:
I think I've dealt with the blockers.

We're going to do this first run in a private Crowdin project as we only want limited translation done for now as we plan to change the content early next year. Once the content is final, everything will move to the same public project as other micro:bit translation work and it will be open for volunteer translation. Subject to change, but that's the working plan.

I've also resolved #330, #375 and #363.

Next step is to add scripts to munge the react-intl string format into one Crowdin will accept. We tested this process previously so it should be straightforward.

microbit-matt-hillsdon:
I have an internal PR for translation sync to Crowdin that I'll test on Monday when there's plenty of time to deal with any fallout.

microbit-matt-hillsdon:
Some toolkit content model things to ponder before finalising the content:
- Perhaps they should all have slugs or ids to help with future URL generation. Tempted to do that now.
- Do we need backlinks? Leaning towards no because we can query quite flexibility with GROQ but worth considering further.

microbit-matt-hillsdon:
I've added a CI-based process for "Explore" for now as it's easy to do and we may make small tweaks in the early new year.
I'll disable it once translation starts. It can also be run locally.

microbit-matt-hillsdon:
I've just added slugs to the toolkit content. I've left the id field in the CMS for the moment for the sake of older branches. I'll remove it once active branches are merged in the next few days (tracked in calendar).

microbit-matt-hillsdon:
Nothing further to do on this issue. I'll check in that the Explore content Crowdin sync runs correctly later today.

Closing.

PR #384, opened 2021-12-20
Basic tooling to convert i18n formats.

microbit-matt-hillsdon:
Import/export with Crowdin is still manual which is fair enough for a
single file.

See #364


lang/en.json
```diff
@@ -571,4 +571,4 @@
     "defaultMessage": "Zoom out",
     "description": "Text label for zoom out button"
   }
-}
+}
```
microbit-matt-hillsdon, line 574: I think this is "sort.js" vs VS Code. Might be nice to fix but not worth worrying about now.

microbit-matt-hillsdon:
I've imported content based on this. I was hoping to update the pseudo translation, but the Crowdin app for it doesn't seem to appear for me. I tried reinstalling it.

Issue #378, opened 2021-12-17
Update split view to use drag handles from the design


microbit-matt-hillsdon:
The latest design uses a clear drag handle on the left-right split view.

We should adopt this and consider whether to use it for the editor-serial split view too.

See the SplitView component and the use of it in Workbench.tsx.

This also needs to cover a way to collapse it entirely via a snap shut approach.

microbit-matt-hillsdon:
The drag handle is more messy than expected. In the design it looks like a split between the toolkit and editor area but it doesn't account for serial. We should discuss in the new year:

https://review-python-editor-next.microbit.org/split-view-drag-handle/

The demo adds a common feature that highlights the line. The drag handle is really awkward though.

microbit-matt-hillsdon:
We now have the ability to collapse the simulator and sidebar via buttons. Nothing further is planned for the moment, though we'd welcome feedback. Closing.

PR #389, opened 2021-12-21
Split view drag handle initial commit

microbit-robert:
I've included the VS Code UI style where a coloured line is drawn over the division when the drag handle is hovered. The positioning of the drag handle vertically is still an issue, particularly with the serial console open.

See #378


src/common/SplitView/SplitView.tsx
```diff
@@ -94,6 +94,8 @@ export const SplitView = ({
         if (size > maximum) {
           size = maximum;
         }
+        // fix offset of dragHandle without affect serial divider
+        size = direction === "column" ? size : size - 15;
```
microbit-robert, line 98: Needs changing from magic number if drag handle is pursued

microbit-matt-hillsdon:
We went with another version of this. Closing.

PR #393, opened 2021-12-22
Ensure we have a complete tree.

### LINKED ISSUE #392: Use state.ensureSyntaxTree not state.syntaxTree when we need the tree to be complete
This is the cause of the occasional test failure in edits.test.ts but also a potential (less likely) real issue in all similar code.

The syntax tree returned from state.syntaxTree may be incomplete (but not out of date). We should switch API where it is essential for correctness.

CC @microbit-robert FYI this is probably the cause of the failure you saw that passed when I re-ran it yesterday.

microbit-matt-hillsdon:
Otherwise, particularly in tests, we can have be missing nodes for the
imports which throws off our calculations.

Closes #392

### DIFF
--- a/src/editor/codemirror/edits.ts
+++ b/src/editor/codemirror/edits.ts
@@ -6,7 +6,7 @@
  * SPDX-License-Identifier: MIT
  */
 import { python } from "@codemirror/lang-python";
-import { syntaxTree } from "@codemirror/language";
+import { ensureSyntaxTree } from "@codemirror/language";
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
@microbit-robert hard to prove this fixes the issue as I never had luck seeing it locally, but it makes sense that it could be the cause and we definitely want to be sure we have the whole tree.

microbit-robert (approved):
LGTM

Issue #329, opened 2021-11-18
[Toolkit] Set image dimensions in portable text to avoid reflow on image load


microbit-matt-hillsdon:
We need to set image dimensions in portable text to avoid reflow on image load. 
The information should be available in the Sanity image asset.

microbit-matt-hillsdon:
See the serializer for `simpleImage` in ToolkitContent.tsx.

PR #394, opened 2021-12-22
Set image height in toolkit

microbit-robert:
Calculate and set image height in toolkit to avoid reflow on image load.

See #329

### DIFF
--- a/src/documentation/ToolkitDocumentation/ToolkitContent.tsx
+++ b/src/documentation/ToolkitDocumentation/ToolkitContent.tsx
@@ -30,7 +30,7 @@ interface ToolkitContentProps {
 
 export const defaultQuality = 80;
 
-export const imageUrlBuilder = unconfiguredImageUrlBuilder()
+const imageUrlBuilder = unconfiguredImageUrlBuilder()
   // Hardcoded for now as there's no practical alternative.
   .projectId("ajwvhvgo")
   .dataset("apps")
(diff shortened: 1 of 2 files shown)


src/documentation/ToolkitDocumentation/ToolkitContent.tsx
```diff
@@ -38,6 +38,27 @@ export const imageUrlBuilder = unconfiguredImageUrlBuilder()
   .dpr(window.devicePixelRatio ?? 1)
   .quality(defaultQuality);
 
+const getImageDimensions = (imageRef: string): number[] | undefined => {
```
microbit-matt-hillsdon, line 41: This will work fine. We also have the option of pulling more information from Sanity which provides some of this in a more convenient form (I actually hadn't realised we needed to add to the query for this, in the Gatsby setup I more often use it's done differently). In that case we'd need to add to the query to resolve the reference to images in each content section and retrieve dimensions.aspectRatio.

microbit-matt-hillsdon: I think it's worth adding a test for an image in ToolkitContent.test.tsx. You can use the same style of [jest snapshot testing](https://jestjs.io/docs/snapshot-testing).

microbit-matt-hillsdon (changes requested):
Let's add a test.

microbit-matt-hillsdon (approved):

PR #395, opened 2021-12-23
Reload toolkit on language change

### LINKED ISSUE #390: [Toolkits] Query for correct language when the language is changed
In ExploreArea we need to depend on the context language (via useSettings) and reload the contents when the query changes.

The top-level document we load is of type `toolkit`. Translations will have a language field set to the same `language` codes we are using in the UI. So we need to parameterise the query by language and either write a clever query or run it twice to do the fallback to English.

Query language reference: https://www.sanity.io/docs/query-cheat-sheet

microbit-robert:
Update query and reload Explore toolkit on changing language setting.

closes #390

### DIFF
--- a/src/documentation/ToolkitDocumentation/api.ts
+++ b/src/documentation/ToolkitDocumentation/api.ts
@@ -8,6 +8,9 @@ import { Toolkit } from "./model";
 // For now we just slurp the whole toolkit at once.
 // Might revisit depending on eventual size.
 const toolkitQuery = (languageId: string): string => {
+  if (!languageId.match(/^[a-z-]+$/g)) {
+    throw new Error("Invalid language id.");
+  }
   return `
   *[_type == "toolkit" && language == "${languageId}" && slug.current == "explore" && !(_id in path("drafts.**"))]{
     id, name, description,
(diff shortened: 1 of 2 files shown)


src/documentation/ToolkitDocumentation/api.ts
```diff
@@ -17,32 +18,42 @@ const toolkitQuery = `
...
-    if (!result) {
-      throw new Error("Unexpected response format");
-    }
-    const toolkits = result as Toolkit[];
-    if (toolkits.length !== 1) {
-      throw new Error("Could not find expected toolkit");
+const fetchToolkitExternal = async (languageId: string): Promise<Toolkit> => {
+  let numRequests = 0;
+  const sanitisedLanguageId = languageId.match(/[a-z]+/g)?.join() ?? "";
+
```
microbit-robert, line 33: Should fallback to English here if sanitisedLanguageId is null?

src/documentation/ToolkitDocumentation/api.ts
```diff
@@ -17,32 +18,42 @@ const toolkitQuery = `
...
+    numRequests++;
+    if (response.ok) {
+      const { result } = await response.json();
+      if (!result) {
+        throw new Error("Unexpected response format");
+      }
+      const toolkits = result as Toolkit[];
+      if (toolkits.length !== 1 && numRequests === 1) {
+        return fetchToolkitInternal("en");
+      }
```
microbit-robert, line 45: This is the neatest way I've found to fall back to English without creating an infinite loop. Open to suggestions if there is a better way.

src/documentation/ToolkitDocumentation/api.ts
```diff
@@ -17,32 +18,42 @@ const toolkitQuery = `
...
-let promise: Promise<Toolkit> | undefined;
+  return fetchToolkitInternal(sanitisedLanguageId);
+};
 
-export const fetchToolkit = async (): Promise<Toolkit> => {
-  if (!promise) {
-    promise = fetchToolkitInternal();
-  }
-  return await promise;
+export const fetchToolkit = async (languageId: string): Promise<Toolkit> => {
```
microbit-robert, line 57: Removed as toolkit isn't fetch again on language change as promise is undefined

src/documentation/ToolkitDocumentation/api.ts
```diff
@@ -17,32 +18,42 @@ const toolkitQuery = `
...
 
-let promise: Promise<Toolkit> | undefined;
+  return fetchToolkitInternal(sanitisedLanguageId);
+};
 
-export const fetchToolkit = async (): Promise<Toolkit> => {
-  if (!promise) {
-    promise = fetchToolkitInternal();
-  }
-  return await promise;
```
microbit-robert, line 47: Removed as toolkit isn't fetch again on language change as promise is undefined

microbit-matt-hillsdon (approved):

Issue #376, opened 2021-12-17
[Toolkits] Switch to two-level design


microbit-matt-hillsdon:
The latest designs use a two-level approach to "Explore" and "Reference" with a "SHOW MORE" / "SHOW LESS" for the detail in each topic entry.

We should switch to this approach with minimal UI change, just switching from "More ->" to "SHOW MORE".

We need to keep the current ability to navigate to the section so probably want to keep the URL behaviour that we have and just use it to control the current expanded section.

Issue #325, opened 2021-11-16
[Toolkits] URL navigation doesn't distinguish between overloads so only one details screen can be seen


microbit-matt-hillsdon:
URL navigation doesn't distinguish between overloads so only one details screen can be seen.

Seen using the work-in-progress microbit.run_every but should apply to existing overloads.

microbit-matt-hillsdon:
We have ids like 1234-1 and 1234-2 and a full name like microbit.display.scroll in both cases.
I think ideally we'd change our id generation code in Pyright to create ids like microbit.display.scroll-2 and use them in place of the full name in the URLs. I don't think it's acceptable to use the more cryptic numeric ids.

Issue #317, opened 2021-11-15
[Toolkits] microbit.audio > (play | is_playing) "More ->" link doesn't work


microbit-matt-hillsdon:
In the reference toolkit, microbit.audio section, the play or is_playing "More ->" link doesn't work. You get sent back to the top of the reference section.

I think it might be audio vs microbit.audio confusion. The third level generally works fine.

microbit-matt-hillsdon:
Since #401 an equivalent issue can be seen when navigating via signature/autocomplete help.

microbit-matt-hillsdon:
Pyright identifies the module as `audio` not `microbit.audio` even if it has been imported via microbit.audio.

In the stubs it's audio.pyi re-exported from microbit/__init__.pyi.

So potentially we can flip that in the stubs or we can cope with the true name differing from how we prefer to display it in the docs.

Stubs PR: https://github.com/microbit-foundation/micropython-microbit-stubs/pull/20

PR #401, opened 2021-12-23
Two level toolkit design with show more/less

microbit-matt-hillsdon:
Maybe separate PR:
- fading background as per design for active one (just add to the effect)
- activation for Explore, no need yet but it's coming in search

See #376


src/documentation/common/ShowMoreButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    color="#262626"
+    variant="unstyled"
+    display="inline-flex"
+    justifyContent="flex-start"
+    size="sm"
+    opacity="0.6"
+    letterSpacing="0px"
+    p="0 12px"
+    textTransform="uppercase"
+    width="max-content"
```
microbit-matt-hillsdon, line 25: To review, feels like there's a lot here.
The font also feels lighter than in XD.

src/documentation/ToolkitDocumentation/ToolkitDocumentation.tsx
```diff
@@ -46,7 +53,7 @@ export const ToolkitDocumentation = ({ toolkit }: ToolkitProps) => {
   );
   return (
     <ActiveTooklitLevel
-      key={urlParam}
+      key={urlParam.length > 0 ? 0 : 1}
```
microbit-matt-hillsdon, line 56: Might be clearer for this to just use `anchor` now.

microbit-matt-hillsdon: Rewriting in terms of anchor here (and surrounding code here and in reference) improved things.

src/documentation/common/ShowMoreButton.tsx
microbit-matt-hillsdon: I've dropped what I can here.
The font weight is the most significant different from the design here.

microbit-matt-hillsdon:
Visually, I still don't like "SHOW MORE" but I think we should probably merge without worrying about that and then consider options (mostly because this is holding up other PRs).

microbit-matt-hillsdon:
Spotted a spacing regression in Reference.
New:
image
Old:
image

microbit-matt-hillsdon:
Preview URL is: 

https://review-python-editor-next.microbit.org/two-level-design-2/

microbit-robert (changes requested):
- Scrolling to reference documentation from the link in the signature help is a little inconsistent. It always scrolls, but not necessarily to the correct place (e.g. for `display.show()`, I can see a 'show less' at the top of the scroll area).
- The link in signature help for anything in `audio` only takes you to references docs level 1, and not into the audio section for some reason.
- If not already inside the relevant second level of the reference toolkit, clicking the link in the signature help a second time after scrolling has finished, appears to cause a flash/glitch in the toolkit. E.g., `display.set_pixel()`

microbit-matt-hillsdon:
There's an issue open about an issue with audio which might be related: #317

microbit-matt-hillsdon:
> * Scrolling to reference documentation from the link in the signature help is a little inconsistent. It always scrolls, but not necessarily to the correct place (e.g. for `display.show()`, I can see a 'show less' at the top of the scroll area).

I think this is #325 in "show more" form. We don't uniquely identify the overload of scroll so we expand both which can cause the offset issue. Can you reproduce this with a method that doesn't have an overload?

> * The link in signature help for anything in `audio` only takes you to references docs level 1, and not into the audio section for some reason.

I'll look at this with #317.

> * If not already inside the relevant second level of the reference toolkit, clicking the link in the signature help a second time after scrolling has finished, appears to cause a flash/glitch in the toolkit. E.g., `display.set_pixel()`

New issue:

URLs are utterly messed up! E.g. ends with `%5Bobject%20Object%5D`. We need to unwrap Anchor.

src/documentation/reference/ReferenceToolkit.tsx
```diff
@@ -3,109 +3,85 @@
...
-      : "back";
+  const [anchor, setAnchor] = useRouterParam(RouterParam.reference);
+  const handleNavigate = useCallback(
+    (id: string | undefined) => {
+      setAnchor(id ? { id } : undefined);
+    },
+    [setAnchor]
+  );
+  const direction = useAnimationDirection(anchor);
+  console.log({ anchor });
```
microbit-matt-hillsdon, line 36: Unwanted debug:

```suggestion
```

microbit-matt-hillsdon:
> > * Scrolling to reference documentation from the link in the signature help is a little inconsistent. It always scrolls, but not necessarily to the correct place (e.g. for `display.show()`, I can see a 'show less' at the top of the scroll area).
> 
> I think this is #325 in "show more" form. We don't uniquely identify the overload of scroll so we expand both which can cause the offset issue. Can you reproduce this with a method that doesn't have an overload?

We could and I've fixed it with a delay so the final height of the container accounts for the show-more expand before the scroll.

> > * The link in signature help for anything in `audio` only takes you to references docs level 1, and not into the audio section for some reason.
> 
> I'll look at this with #317.

Still leaving that to look at separately.

> > * If not already inside the relevant second level of the reference toolkit, clicking the link in the signature help a second time after scrolling has finished, appears to cause a flash/glitch in the toolkit. E.g., `display.set_pixel()`
> 
> I can't make this happen.

We reproduced this and I've finally fixed it in Slide.tsx (unintentional tree structure change causing component recreation).

> New issue:
> 
> URLs are utterly messed up! E.g. ends with `%5Bobject%20Object%5D`. We need to unwrap Anchor.

This is also fixed.

I think this PR is good to go. I'll address the fading background separately.

microbit-matt-hillsdon:
> Spotted a spacing regression in Reference. New: image Old: image

This is also fixed.

microbit-matt-hillsdon:
One issue to look at after merge:

- If there's a CM in the Collapse then it seems to fade out but not slide up, e.g. display.scroll. Clearer if you remove the other content in the Collapse. What's going on?

microbit-robert (approved):
LGTM

microbit-matt-hillsdon:
> * If there's a CM in the Collapse then it seems to fade out but not slide up, e.g. display.scroll. Clearer if you remove the other content in the Collapse

Now #407

PR #404, opened 2022-01-04
Update tabs to match design

### LINKED ISSUE #398: Align tabs with the new designs
image

- [ ] Review locations of tabs and align
- [ ] Change colour of tabs to the off-white and muted Files tab.
- Keep "Explore" as "Explore"
- Keep the underline focus indicator 
- Probably keep tab internal sizing/font sizes but review

microbit-robert:
Align the Files position and colors with the design.

We've intentionally left the spacing, sizing and corners.

Closes #398

### DIFF
--- a/src/workbench/SideBar.tsx
+++ b/src/workbench/SideBar.tsx
@@ -33,7 +33,19 @@ import SettingsMenu from "../settings/SettingsMenu";
 import HelpMenu from "./HelpMenu";
 import ReleaseDialogs from "./ReleaseDialogs";
 import ReleaseNotice, { useReleaseDialogState } from "./ReleaseNotice";
-import CustomTab from "./CustomTab";
+import SideBarTab from "./SideBarTab";
+
+export const cornerSize = 32;
+
+export interface Pane {
+  id: string;
+  icon: IconType;
+  title: string;
+  nav?: ReactNode;
+  contents: ReactNode;
+  color: string;
+  mb?: string;
+}
 
 interface SideBarProps extends BoxProps {
   selectedFile: string | undefined;
@@ -50,6 +62,8 @@ const SideBar = ({
   ...props
 }: SideBarProps) => {
   const intl = useIntl();
+  const [releaseDialog, setReleaseDialog] = useReleaseDialogState();
+  const brand = useDeployment();
   const panes: Pane[] = useMemo(() => {
     const result = [
       {
@@ -57,6 +71,7 @@ const SideBar = ({
         title: intl.formatMessage({ id: "explore-tab" }),
         icon: PythonLogo as IconType,
         contents: <ExploreArea />,
+        color: "gray.50",
       },
       {
         id: "reference",
@@ -65,6 +80,8 @@ const SideBar = ({
         // in documentation.ts (used for CM documentation tooltips).
         icon: VscLibrary,
         contents: <ReferenceArea />,
+        color: "gray.50",
+        mb: "auto",
       },
       {
         id: "files",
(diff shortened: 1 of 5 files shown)


src/workbench/SideBar.tsx
```diff
@@ -161,23 +161,21 @@ const SideBarContents = ({ panes, ...props }: SideBarContentsProps) => {
...
               height={width}
               width={width}
               p={0}
               position="relative"
               className="sidebar-tab" // Used for custom outline below
               onClick={handleTabClick}
+              borderRadius={`${cornerSize}px 0 0 ${cornerSize}px`} //temp override - apply to branding
+              pr="6px" //temp override - apply to branding
+              mb={p.id === "reference" ? "auto" : 3}
+              mt={p.id === "files" ? 3 : 0}
```
microbit-matt-hillsdon, line 178: Not a fan of how this switches on id. I'd rather this information was at the level of the parent rather than a child switching on its identity.

I think we should either move this information to `p` (e.g. add a color field etc) or rejig so we're creating a component instance for each tab that takes the information as props and remove the `panes` approach entirely.

src/workbench/SideBar.tsx
```diff
@@ -244,25 +242,36 @@ const SideBarContents = ({ panes, ...props }: SideBarContentsProps) => {
...
+    {...props}
+    pointerEvents="none"
+    width={`${cornerSize}px`}
+    height={`${cornerSize}px`}
+  >
     <svg
       width="100%"
       height="100%"
-      viewBox="0 0 32 32"
+      viewBox={`0 0 ${cornerSize} ${cornerSize}`}
```
microbit-matt-hillsdon, line 254: Thanks, this is better. I don't particularly like this corner approach, so better ideas very welcome if you have them (separate PR).

microbit-matt-hillsdon:
I think this looks good. We'll chat about the theme split when we catch-up.

Please would you share this for internal feedback on how the separate Files tab and reduced font size for the tabs feels to folks?

src/workbench/SideBar.tsx
```diff
@@ -161,23 +161,21 @@ const SideBarContents = ({ panes, ...props }: SideBarContentsProps) => {
...
-            ml="auto"
-            mr="auto"
-            mb="max(11.5vh, 7.7rem)"
-          ></Box>
+          <Box width="3.75rem" flex={1} maxHeight="8.9rem" minHeight={7}></Box>
           {panes.map((p, i) => (
             <Tab
               key={p.id}
-              color="white"
+              color={p.id === "files" ? "gray.50" : "#F5F6F8"}
```
microbit-matt-hillsdon, line 168: Color with me to add to theme or change to match existing.

src/workbench/SideBar.tsx
microbit-matt-hillsdon: Agreed to revert the corners but good to keep the constant.

Not perfect that the theme also has a matching constant but we can revisit another time.

src/workbench/SideBar.tsx
```diff
@@ -161,62 +157,27 @@ const SideBarContents = ({ panes, ...props }: SideBarContentsProps) => {
...
-                </VStack>
-              </VStack>
-            </Tab>
-          ))}
-          <VStack mt="auto" mb={1} spacing={0.5} color="white">
+          <Box width="3.75rem" flex={1} maxHeight="8.9rem" minHeight={8}></Box>
+          <CustomTab
+            color="gray.50"
+            handleTabClick={handleTabClick}
+            activeTabId={tab ? tab : panes[0].id}
```
microbit-robert, line 164: `tab` is undefined by default. This feels like a fudge to have the 'explore' tab the default tab on page load after the refactor. Better way?

microbit-matt-hillsdon (approved):

PR #408, opened 2022-01-06
Update to version that moves audio definitions.

### LINKED ISSUE #317: [Toolkits] microbit.audio > (play | is_playing) "More ->" link doesn't work

microbit-matt-hillsdon:
Closes #317


microbit-matt-hillsdon:
Interestingly this has regressed completions in microbit.audio. Importing audio itself is fine.

The pin0, MicroBitDigitalPin and Sound imports that aren't re-exported from audio appear as if they are available on microbit.audio. This is unexpected to me at least. The reverse wasn't an issue previously (i.e. via import audio has no audio.pin0).

This is a preexisting issue for all the modules already similarly structured to audio so I think it makes sense to merge this and fix separately. I'm not clear on whether it's a Pyright issue or a stubs issue so it'll need some research.

It's the reason there's display.Image which is probably the easiest issue to encounter.

microbit-matt-hillsdon:
Raised #410 for reexport issue. Merging.

PR #416, opened 2022-01-10
Highlight activated reference entry

microbit-matt-hillsdon:
This is based on the design in Search which applies to Explore not Reference so adapted a little.

I've stopped using ToolkitListItem here as I needed to significantly change the spacing as I needed padding inside of the highlight rather than margin around it. I think perhaps it should only be used for the top level list in Explore now and therefore can probably be removed in time.

To be honest, I'm not sure the second level should be a list at all, what we have is content in sections with titles. I'll review that separately.

It makes the problem with overrides much more obvious (e.g. display.show) but I think that's just incentive to fix it soon, not a blocker for this.

See #376

### DIFF
--- a/src/documentation/ToolkitDocumentation/Highlight.tsx
+++ b/src/documentation/ToolkitDocumentation/Highlight.tsx
@@ -6,8 +6,11 @@ interface HighlightProps extends BoxProps {
 
 const Highlight = ({ value, children, ...props }: HighlightProps) => {
   const style = value
-    ? { backgroundColor: "brand.100" }
-    : { transition: "background-color ease-in 0.5s" };
+    ? {
+        backgroundColor: "brand.100",
+        transition: "background-color ease-out 0.2s",
+      }
+    : { transition: "background-color ease-in 0.6s" };
   return (
     <Box {...props} {...style} borderLeftRadius="md">
       {children}
--- a/src/documentation/ToolkitDocumentation/Highlight.tsx
+++ b/src/documentation/ToolkitDocumentation/Highlight.tsx
@@ -6,10 +6,10 @@ interface HighlightProps extends BoxProps {
 
 const Highlight = ({ value, children, ...props }: HighlightProps) => {
   const style = value
-    ? { backgroundColor: "blue.200" }
+    ? { backgroundColor: "#e1dbF3" }
     : { transition: "background-color ease-in 0.5s" };
   return (
-    <Box {...props} {...style}>
+    <Box {...props} {...style} borderLeftRadius="md">
       {children}
     </Box>
   );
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
We'll do something equivalent for Explore when we implement search.

src/documentation/ToolkitDocumentation/Highlight.tsx
```diff
@@ -0,0 +1,18 @@
+import { Box, BoxProps } from "@chakra-ui/react";
+
+interface HighlightProps extends BoxProps {
+  value: boolean;
+}
+
+const Highlight = ({ value, children, ...props }: HighlightProps) => {
+  const style = value
+    ? { backgroundColor: "#e1dbF3" }
```
microbit-matt-hillsdon, line 9: Color needs moving to theme.

microbit-matt-hillsdon: Done.

src/documentation/reference/ReferenceNode.tsx
```diff
@@ -79,6 +81,12 @@ const ReferenceNode = ({
             behavior: prefersReducedMotion ? "auto" : "smooth",
           });
         }
+        setTimeout(() => {
+          setHighlighting(true);
+          setTimeout(() => {
+            setHighlighting(false);
+          }, 3000);
+        }, 300);
       }, 150);
```
microbit-matt-hillsdon, line 90: Let me know what you think about timings here. The design only had the short fade so far as I can tell, but we need to have it linger for a while to make the initial selection clear. 

Or perhaps it's supposed to remain highlighted until you interact with the toolbar? What do you think?

microbit-matt-hillsdon: Added a fade-in after discussion.

@microbit-robert, if this is otherwise OK, then I think we should review and merge as is, then I'll tweak based on feedback once we've heard from more folk.

src/documentation/reference/ReferenceNode.tsx
```diff
@@ -91,27 +99,40 @@ const ReferenceNode = ({
...
-      <ReferenceNodeChildren docs={docs} anchor={anchor} />
+      <Highlight
+        onClick={handleHighlightClick}
+        value={highlighting}
+        mt={1}
+        mb={1}
+        pl={3}
+        pr={3}
+        pt={3}
+        pb={3}
```
microbit-robert, line 127: Just `p={3}`?

microbit-robert (approved):
LGTM, one optional change on comment

src/documentation/reference/ReferenceNode.tsx
microbit-matt-hillsdon: Hah, yes, they started different and I hadn't noticed they'd ended up the same. Thanks.

microbit-matt-hillsdon: 85f378547cef568cbf8d3e3d5401cbab3bdb341f

PR #418, opened 2022-01-10
Show "ghost code" during drag and drop

microbit-matt-hillsdon:
The colours and fade are still being discussed but are in the right ballpark.

We support only a single drag or drop, i.e. dragging again will clear the existing decorations. The highlighted text can be non-contiguous though (consider imports). 

We could consider changing this but I think it'd be more confusing than helpful to have multigenerational ghosts.

### DIFF
--- a/src/editor/codemirror/dnd-decorations.test.ts
+++ b/src/editor/codemirror/dnd-decorations.test.ts
@@ -157,7 +157,7 @@ const createView = (doc: Text = Text.of([""])): EditorView => {
 };
 
 /**
- * Note that, unlike normal CodeMirror usage, should use the update's
+ * Note that, unlike normal CodeMirror usage, you must use the update's
  * view for any subsequent changes.
  */
 const createViewUpdate = (
--- a/src/editor/codemirror/dnd-decorations.test.ts
+++ b/src/editor/codemirror/dnd-decorations.test.ts
@@ -1,6 +1,5 @@
-import { isInputEvent } from "@chakra-ui/utils";
 import { ChangeSet, EditorState, Text, Transaction } from "@codemirror/state";
-import { DecorationSet, EditorView, ViewUpdate } from "@codemirror/view";
+import { EditorView, ViewUpdate } from "@codemirror/view";
 import { DndDecorationsViewPlugin } from "./dnd-decorations";
 
 describe("dndDecorations", () => {
@@ -157,6 +156,10 @@ const createView = (doc: Text = Text.of([""])): EditorView => {
   } as Partial<EditorView> as unknown as EditorView;
 };
 
+/**
+ * Note that, unlike normal CodeMirror usage, should use the update's
+ * view for any subsequent changes.
+ */
 const createViewUpdate = (
   view: EditorView,
   docChanged: boolean,
@@ -164,14 +167,10 @@ const createViewUpdate = (
 ): ViewUpdate => {
   const state = transaction ? transaction.state || view.state : view.state;
   return {
-    view: {
-      ...view,
-      state,
-      visibleRanges: [{ from: 0, to: state.doc.length - 1 }],
-    },
+    view: createView(state.doc),
     state,
     transactions: transaction ? [transaction] : [],
     changes: transaction ? transaction.changes : ChangeSet.empty(0),
     docChanged,
-  } as unknown as any;
+  } as Partial<ViewUpdate> as unknown as any;
 };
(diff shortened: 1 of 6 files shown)


src/editor/codemirror/dnd.ts
```diff
@@ -15,6 +23,77 @@ export const debug = (message: string, ...args: any) => {
...
+
+    constructor(view: EditorView) {
+      this.decorations = ghostDecorations(view, new Set([]));
+    }
+
+    update(update: ViewUpdate) {
+      if (update.docChanged) {
+        const ghosts =
+          update.transactions.length === 1 &&
+          update.transactions[0].annotation(Transaction.addToHistory) === false;
```
microbit-matt-hillsdon, line 46: Maybe some other annotation on the transaction would be more appropriate.

src/editor/codemirror/dnd-decorations.ts
```diff
@@ -0,0 +1,120 @@
...
+        ) {
+          this.droppedDoneLineFroms.clear();
+          this.droppedRecentLineFroms.forEach((v) =>
+            this.droppedDoneLineFroms.add(v)
+          );
+          this.previewLineFroms.clear();
+          this.droppedRecentLineFroms.clear();
+        }
+      }
+
```
microbit-matt-hillsdon, line 37: This is a bit messy and definitely needs testing.
Could we have an array sized to the visible lines or some other better structure?

src/editor/codemirror/dnd.ts
```diff
@@ -66,17 +67,13 @@ const clearSuppressChildDragEnterLeave = (view: EditorView) => {
...
- * Note this requires coordination from the drag end via {@link setDraggedCode}.
- */
-export const dndSupport = () => {
+const dndHandlers = () => {
   let lastDragPos: LastDragPos | undefined;
 
   const revertPreview = (view: EditorView) => {
     if (lastDragPos) {
       view.dispatch({
+        userEvent: "dnd.cleanup",
```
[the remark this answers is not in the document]
microbit-matt-hillsdon: I think I plan to merge this one then fix up the other one. Either of them could use another annotation or we can add to the hierarchy.

src/editor/codemirror/dnd-decorations.ts
microbit-matt-hillsdon: We can't reasonably do this but I've tidied up a bit while keeping the approach essentially the same.

src/editor/codemirror/dnd-decorations.ts
```diff
@@ -0,0 +1,115 @@
...
+  EditorView.theme({
+    ".cm-preview": {
+      backgroundColor: "#fefcbfdd",
+    },
+    ".cm-dropped--recent": {
+      backgroundColor: "#fefcbf99",
+    },
+    ".cm-dropped--done": {
+      transition: "background-color ease-in 1s",
+    },
```
microbit-matt-hillsdon, line 110: We're still chatting about the details of the styling but probably OK to merge and continue the debate.

src/editor/codemirror/dnd.ts
```diff
@@ -172,3 +170,10 @@ export const dndSupport = () => {
     }),
   ];
 };
+
+/**
+ * Support for dropping code snippets.
+ *
+ * Note this requires coordination from the drag end via {@link setDraggedCode}.
```
microbit-matt-hillsdon, line 177: This is out of date.

microbit-matt-hillsdon: Oh no, it will be once the reference dnd branch is merged. Not one for this PR.

src/editor/codemirror/dnd-decorations.test.ts
```diff
@@ -0,0 +1,176 @@
...
+  return {
+    visibleRanges: [{ from: 0, to: doc.length - 1 }],
+    state: EditorState.create({ doc }),
+    dispatch: jest.fn(),
+  } as Partial<EditorView> as unknown as EditorView;
+};
+
+/**
+ * Note that, unlike normal CodeMirror usage, should use the update's
+ * view for any subsequent changes.
```
microbit-matt-hillsdon, line 161:
```suggestion
 * Note that, unlike normal CodeMirror usage, you must use the update's
 * view for any subsequent changes.
```

microbit-robert (approved):
LGTM

src/editor/codemirror/dnd-decorations.ts
microbit-matt-hillsdon: We agreed new values. See 582bf5937bc84ce6814c67848913364d1079f7d1

src/editor/codemirror/dnd.ts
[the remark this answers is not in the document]
microbit-matt-hillsdon: It worked out OK with e.g. dnd.drop.call vs dnd.drop.explore. If there was reason to we could move the call/explore distinction to a separate transaction annotation.

PR #422, opened 2022-01-12
Fix top bar order

### LINKED ISSUE #406: Fix top bar order when viewing non-main file
image

microbit-robert:
Fix top bar order when viewing non-main file.

Closes #406


microbit-matt-hillsdon (approved):
Looks great, thanks.

PR #429, opened 2022-01-14
Handle overrides with unique ids.

### LINKED ISSUE #325: [Toolkits] URL navigation doesn't distinguish between overloads so only one details screen can be seen

microbit-matt-hillsdon:
Closes #325


src/documentation/ReferenceArea.tsx
```diff
@@ -19,6 +19,7 @@ const ReferenceArea = () => {
         await client.initialize();
         const docs = await apiDocs(client);
         pullModulesToTop(docs);
+        console.log(JSON.stringify(docs, null, 2));
```
microbit-robert, line 22: Remove debugging log

microbit-matt-hillsdon, line 22:
```suggestion
```

microbit-robert (approved):
LGTM

PR #433, opened 2022-01-17
Remove unhelpful alt text.

microbit-matt-hillsdon:
The text we already have is sufficient; the icons are just decorative
(or will be, when we add them).


microbit-robert (changes requested):
This also needs removing from ToolkitBreadcrumbHeading.tsx

microbit-matt-hillsdon:
> This also needs removing from ToolkitBreadcrumbHeading.tsx

186164c11f629becf4dc7badf2c497b1cc6a0b07 (squashed)

microbit-robert (approved):

PR #443, opened 2022-01-19
Add V2 tag element

### LINKED ISSUE #439: [Toolkits] Switch Explore (V2) suffix to use tag shown in design
The implementation uses a bracketed suffix but the design shows a tag:

image

We should probably use Chakra's Tag component.

See calls to `isV2Only`.

microbit-robert:
Add V2 suffix as a tag as shown in the design.

Closes #439

### DIFF
--- a/src/documentation/explore/ToolkitTopLevelListItem.tsx
+++ b/src/documentation/explore/ToolkitTopLevelListItem.tsx
@@ -15,6 +15,7 @@ import {
   ListItemProps,
   ListItem,
   Divider,
+  Tag,
 } from "@chakra-ui/react";
 import { imageUrlBuilder } from "../../common/imageUrlBuilder";
 
@@ -22,6 +23,7 @@ interface ToolkitTopLevelListItemProps {
   name: string;
   description: ReactNode;
   icon?: ToolkitImage;
+  isV2Only?: Boolean;
   onForward: () => void;
 }
 
@@ -29,6 +31,7 @@ const ToolkitTopLevelListItem = ({
   name,
   description,
   icon,
+  isV2Only,
   onForward,
 }: ToolkitTopLevelListItemProps) => {
   const intl = useIntl();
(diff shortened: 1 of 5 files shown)


microbit-robert:
Minor deviations from the design are noted below. These values don't fit well with the chakra theme by default, and the font size looks nicer at 14px regardless. 

- Tag font size is 14px instead of 15px in the design.
- Tag border radius is 6px instead of 4px in the design.

microbit-robert:
Failed build due to e2e test. Not seen this one fail before:
image

src/documentation/explore/ToolkitTopLevelListItem.tsx
```diff
@@ -15,20 +15,23 @@ import {
...
   Divider,
+  Tag,
 } from "@chakra-ui/react";
 import { imageUrlBuilder } from "../../common/imageUrlBuilder";
 
 interface ToolkitTopLevelListItemProps {
   name: string;
   description: ReactNode;
   icon?: ToolkitImage;
+  isV2Only?: Boolean;
```
microbit-matt-hillsdon, line 26:
```suggestion
  isV2Only?: boolean;
```

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object

src/documentation/explore/ToolkitTopLevelListItem.tsx
```diff
@@ -41,9 +44,17 @@ const ToolkitTopLevelListItem = ({
...
+          <HStack>
+            <Text as="h3" fontSize="lg" fontWeight="semibold">
+              {name}
+            </Text>
+            {isV2Only && (
+              <Tag fontWeight="semibold" background="brand.500" color="gray.25">
+                V2
+              </Tag>
+            )}
+          </HStack>
```
microbit-matt-hillsdon, line 56: Maybe this should be a component?

microbit-robert: [`8cc14ff2ff`](https://github.com/microbit-foundation/python-editor-next/pull/443/commits/8cc14ff2fff445b903115f94c87ea845d7f71e49)

src/documentation/explore/ToolkitTopLevelListItem.tsx
microbit-robert: [`8cc14ff2ff`](https://github.com/microbit-foundation/python-editor-next/pull/443/commits/8cc14ff2fff445b903115f94c87ea845d7f71e49)

microbit-matt-hillsdon (approved):
Looks good, thanks.

PR #456, opened 2022-01-21
Disable grammarly differently.

microbit-matt-hillsdon:
I'll get someone who uses it to compare.


microbit-matt-hillsdon:
Closing without merging. I cannot get Grammarly's Chrome extension to break anything with things as they are. It's also really easy to disable on a per site basis. Closing, will revisit if we get user reports.

Issue #440, opened 2022-01-18
[Search] Initial implementation


microbit-matt-hillsdon:
## Description

Add offline-capable search matching the UI design behind a "search" feature flag.

We've opted to try lunr as an offline search library with reasonably wide language support. elasticlunr is another option that shares the same language support. 

The plan is to integrate lunr or similar providing a basic search API via a hook and then add the search UI from the design, perhaps omitting more involved features in this first pass.

## Remaining work

Must for user testing
- [x] Add French content and then update lunr initialization to respect language. Deal with code splitting etc.
- [x] Prune common coding terms from the English stop words and see whether that helps. E.g. currently can't search for if/while/else
- [x] Review keyboard navigation
    - Hover highlights as part of this
    - Add a href to the links for tabbing (serialise router state), add an onClick to the thing as a whole
- [x] Translatable strings - also need to consider fallback
- #490
    - Show more context? Show fewer extracts if there are many matches? Don't truncate words? Try to equalise the text length shown?
- [x] Tracking of things you've clicked through to (with the lighter highlight)
- #489

Deferrable?
-  Decide whether we do 100+ results or exact number, if so what are the full rules.
    - The performance is fine for the moment. Revisit when we remove some English stop words.
- Add narrower search button when the sidebar is narrower
    - Rolling in with general tablet work.
- Filter by type
    - Raised #481 
- Recent searches
    - Raised #482 
- Highlight terms after you click through (not sure I want to do this at all)
    - I just don't think we should do this.

## Design changes/issues

- The search box that's initially visible on larger screens is odd. Is it an input? If so we'll need to move the focus when you click on it. Is it a button that looks like an input? If so that's very confusing. Perhaps it should just be a button as per the smaller width design?
    - For now I've matched the design but I'm unsure about it.

- We want an Explore / Reference split as shown in the design (though with Explore/Reference not the text used in the design).

- We might want to make the toolkit and topic clearer as the text is muted in the design.

- Padding is a bit odd in the results pane. Fine in our implementation.

microbit-matt-hillsdon:
The serialization of the Sanity "portable text" for indexing can be handled with something like this: 
https://www.sanity.io/schemas/portable-text-to-plain-text-cc845843

microbit-matt-hillsdon:
[...]

PR #459, opened 2022-01-24
Search top level

microbit-robert:
Adds the ability to search the top level (topics for the explore content and modules for the reference content). The explore topic introduction text is also searchable, and links to the topic as opposed to any specific topic entry.

See #440


microbit-matt-hillsdon (approved):
Jonny expressed some doubt in discussion this morning, but I think it's really natural to be able to search for the words you've seen at the top-level too. You can also use it as a quick way to navigate to a topic (once we have keyboard nav). Let's get it in and I'll flag it with him once the search feature as a whole is more rounded out.

PR #460, opened 2022-01-24
Handle the no match case.

microbit-matt-hillsdon:
Previously returned truncated first/last/only text extract.


microbit-robert (approved):
LGTM

PR #461, opened 2022-01-24
Search persistence UI

microbit-robert:
Add UI to show existing search query in header. 

See #440

### DIFF
--- a/src/documentation/Search.tsx
+++ b/src/documentation/Search.tsx
@@ -21,7 +21,7 @@ interface SearchProps {
   query: string;
   setQuery: React.Dispatch<React.SetStateAction<string>>;
   onQueryChange: React.ChangeEventHandler<HTMLInputElement>;
-  onClear: React.MouseEventHandler<HTMLButtonElement>;
+  onClear: (inputRef: React.RefObject<HTMLInputElement>) => void;
 }
 
 const Search = ({
@@ -67,7 +67,7 @@ const Search = ({
               // Also used for Zoom, move to theme.
               color="#838383"
               icon={<RiCloseLine />}
-              onClick={onClear}
+              onClick={() => onClear(ref)}
             />
           </InputRightElement>
         </InputGroup>
--- a/src/documentation/Search.tsx
+++ b/src/documentation/Search.tsx
@@ -9,7 +9,7 @@ import {
   Stack,
   Text,
 } from "@chakra-ui/react";
-import { useRef } from "react";
+import { useCallback, useRef } from "react";
 import { RiCloseLine, RiSearch2Line } from "react-icons/ri";
 import { SearchResults } from "./search-hooks";
 import SearchResultList from "./SearchResultList";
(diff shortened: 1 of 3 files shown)


src/workbench/SideBarHeader.tsx
```diff
@@ -42,15 +44,26 @@ const SideBarHeader = () => {
       },
       [search]
     );
+  const handleClear = useCallback(() => {
+    setQuery("");
+    setResults(undefined);
+    if (ref.current) {
+      ref.current.focus();
```
microbit-matt-hillsdon, line 51: This isn't focusing the same thing it was before.

src/documentation/Search.tsx
```diff
@@ -30,15 +31,9 @@ const Search = ({
...
+  onClear,
 }: SearchProps) => {
   const ref = useRef<HTMLInputElement>(null);
-  const handleClear = useCallback(() => {
-    setQuery("");
-    setResults(undefined);
-    if (ref.current) {
-      ref.current.focus();
-    }
-  }, [setQuery, setResults]);
```
microbit-matt-hillsdon, line 41: Let's try keeping this to do the focus handling then call onClear for the parent state to update.

src/workbench/SideBarHeader.tsx
```diff
@@ -133,6 +147,42 @@ const SideBarHeader = () => {
             Search
           </Button>
         )}
+        {flags.search && query && (
+          <ButtonGroup
+            isAttached
+            backgroundColor="white"
+            width="full"
+            borderRadius="3xl"
```
microbit-matt-hillsdon, line 155: The focus outline looks very odd. Let's try a pattern similar to https://chakra-ui.com/docs/form/input so the focus ring is around the whole element and the other button is visually inside (via absolute positioning).

microbit-matt-hillsdon (changes requested):
Minor tweaks but looks good.

src/workbench/SideBarHeader.tsx
```diff
@@ -48,15 +50,29 @@ const SideBarHeader = () => {
...
-    ? ref.current!.clientWidth - offset - 14 + "px"
+  const width = sidebarWidth.current
+    ? sidebarWidth.current!.clientWidth - offset - 14 + "px"
     : undefined;
+
+  const { sizedPaneSize } = useSplitViewContext();
+  const [searchWidth, setSearchWidth] = useState<number | undefined>();
+  useEffect(() => {
+    setSearchWidth((sizedPaneSize || 0) - offset - 14);
+  }, [offset, sizedPaneSize]);
```
microbit-robert, line 70: We need a better way than this. It causes so much lag on moving the divider.

The problem I was trying to solve was having a long search query when the search dialog is closed. Avoid text or it's container spilling over into the editor. Avoid this happening when the user moves the split view divider. I think we need a `maxWidth` from somewhere. Putting `overflow="hidden"` in the wrong place also hides the focus ring.

microbit-robert: I think I just pushed a fix for it, which was simple, and I was overcomplicating it.

microbit-matt-hillsdon (approved):

PR #462, opened 2022-01-24
Move search indexing to a worker

microbit-matt-hillsdon:
Moves code around so the lunr code is only depended on from the worker and the worker doesn't pull in any code it doesn't need.


src/documentation/search/extracts.ts
```diff
@@ -1,14 +1,13 @@
...
+import { Extract } from "./common";
 
 export type Position = [number, number];
 
-const sortByStart = (positions: Position[]): Position[] =>
-  sortBy(positions, (p) => p[0]);
+// Avoid lodash in the worker
+export const sortByStart = (positions: Position[]): Position[] => {
+  const copy = [...positions];
+  copy.sort((a, b) => (a[0] > b[0] ? 1 : a[0] < b[0] ? -1 : 0));
```
microbit-matt-hillsdon, line 8: Fairly sure it started off life as this in your implementation!

src/documentation/search/search-hooks.tsx
```diff
@@ -0,0 +1,40 @@
...
+  if (!value) {
+    throw new Error("Missing provider!");
+  }
+  return value;
+};
+
+const SearchProvider = ({ children }: { children: ReactNode }) => {
+  const { exploreToolkit, referenceToolkit } = useToolkitState();
+  useEffect(() => {
+    // Wait for both, we'll be re-run for each at start-up.
```
microbit-robert, line 29: `we'll` is a typo?

src/documentation/search/extracts.test.ts
```diff
@@ -62,3 +62,19 @@ describe("contextExtracts", () => {
     ]);
   });
 });
+
+describe("sortByStart", () => {
+  it("works", () => {
```
microbit-robert, line 67: Should we provide a better name for this test?

microbit-matt-hillsdon: 9d67c7f

src/documentation/search/search-hooks.tsx
microbit-matt-hillsdon: I was identifying with the effect... for some reason. Replaced with something clearer in d8a586a92baaac98b8d17256dc75204ab16c5e28.

microbit-robert (approved):
LGTM

microbit-matt-hillsdon:
Merged main and fixed conflicts.

PR #452, opened 2022-01-20
Initial feature-flagged search 

microbit-matt-hillsdon:
First PR for #440.

There's lots still to do tracked on the issue.

PR #464, opened 2022-01-24
Language aware search (just French for now)

microbit-matt-hillsdon:
You can see that the French support is split out into build/static/js/1.67d6d668.chunk.worker.js and is loaded dynamically, so this approach will scale as we add languages.

Depends on #479 (review/merge that first, then I'll rebase for easier review here)

See #452

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13632,6 +13632,11 @@
       "resolved": "https://registry.npmjs.org/lunr/-/lunr-2.3.9.tgz",
       "integrity": "sha512-zTU3DaZaF3Rt9rhN3uBMGQD3dD2/vFQqnvZCDv4dl5iOzq2IZQqTxu90r4E5J+nP70J3ilqVCrbho2eWaeW8Ow=="
     },
+    "lunr-languages": {
+      "version": "1.9.0",
+      "resolved": "https://registry.npmjs.org/lunr-languages/-/lunr-languages-1.9.0.tgz",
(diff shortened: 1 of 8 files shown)


microbit-matt-hillsdon:
In addition to (or perhaps the same as?) the e2e failure I can trigger an error with an immediate search.

microbit-robert:
> @microbit-robert I'm puzzling over why this has broken the serarch e2e test, but any other feedback welcome.

Not sure on the e2e test. No results loaded? 

I get this error in the console when switching the language to French, and search then fails. It works when switching back to English.
image

microbit-matt-hillsdon:
> > @microbit-robert I'm puzzling over why this has broken the serarch e2e test, but any other feedback welcome.
> 
> Not sure on the e2e test. No results loaded?

I think removing the reindexing has broken the blocking on first index. I will reinstate, just need to find a way to test now it's async.

> I get this error in the console when switching the language to French, and search then fails. It works when switching back to English. image

Interesting, the path is broken, with static/js twice.

https://review-python-editor-next.microbit.org/lunr-languages/static/js/static/js/1.67d6d668.chunk.worker.js

microbit-matt-hillsdon:
Hopeful that I've fixed e2e but the other issue seems more worrying as it's more like a webpack bug.

microbit-matt-hillsdon:
I've merged the change from #479 in to this branch which should fix the deployment issue that meant the French lunr plugin isn't loaded.

microbit-matt-hillsdon:
@microbit-robert, as you said in chat, we need to fix this to clear the search when the language changes. I'd like to tackle this separately, once all current PRs are in. I think it's time to pull the search state up to the provider, then we can naturally clear it when we reindex.

microbit-robert (approved):
Looks good. All issues appear to be resolved.

PR #466, opened 2022-01-24
Avoid indexing topics for singleton topics.

### LINKED ISSUE #465: [Search] Odd extracts
Mixed in with otherwise sensible results:

image

image

I think these are topic matches but they're using some other string for the extracts.

microbit-matt-hillsdon:
Closes #465 

We'll revisit their ids and visual treatment when we can more easily edit the content without translation issues.


microbit-robert (approved):
LGTM

PR #468, opened 2022-01-25
Disable flags for e2e tests.

microbit-matt-hillsdon:
Add system to opt into specific flags.
Default the noWelcome flag as otherwise every test must deal with the dialog.

The system to pass flags is currently unused, but I'll add a search test soon to
prove it works.

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -12,7 +12,7 @@ import * as os from "os";
 import * as path from "path";
 import "pptr-testing-library/extend";
 import puppeteer, { Browser, Dialog, ElementHandle, Page } from "puppeteer";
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
This change is motivated by the issue fixed in #467 where e2e tests passed on the branch only because the search flag was enabled but the app was broken on on main. We were lucky that the e2e tests on main caught it at all: it was a fluke due to the URL used for the migration test accidentally the flag from * to */.

We care much more about the app working with flags disabled than with flags enabled as the latter only impacts users opting in to work-in-progress whereas the former affects everyone.

microbit-robert (approved):
LGTM

PR #469, opened 2022-01-25
Update script for 2022 and update headers.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
@microbit-robert, we should be manually updating 2021 files to say " - 2022" on edit, when we fail at that I'll write a cleverer script!

microbit-matt-hillsdon:
No need to review, just FYI.

PR #471, opened 2022-01-25
Workaround for autocomplete e2e issue.

### LINKED ISSUE #419: Flakey e2e test
```
FAIL  src/e2e/autocomplete.test.ts (18.068 s)
  ● Browser - autocomplete and signature help tests › does not insert brackets for import completion

    Evaluation failed: Error: Unable to find role="option"
```

microbit-matt-hillsdon:
Optimistically closes #419

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -396,7 +396,10 @@ export class App {
    */
   async typeInEditor(text: string): Promise<void> {
     const content = await this.focusEditorContent();
-    return content.type(text);
+    // The short delay seems to improve reliability triggering autocomplete.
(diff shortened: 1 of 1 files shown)


src/e2e/app.ts
```diff
@@ -685,7 +688,7 @@ export class App {
       },
       defaultWaitForOptions
     );
-    option.click();
+    await option.click();
```
microbit-matt-hillsdon, line 691: This is a bug but not the cause of our issue as it's after the failure.

microbit-robert (approved):

PR #472, opened 2022-01-25
Add keyboard shortcut for search

microbit-robert:
Adds ability to handle keyboard shortcuts and implements shortcut for search.

See #440


microbit-matt-hillsdon:
Went with a simpler approach for now on #473 but we will want to pull up in future, see inline comments.

PR #473, opened 2022-01-25
Simple keyboard handling with a hint of the future

microbit-matt-hillsdon:


src/workbench/SideBarHeader.tsx
```diff
@@ -37,6 +37,27 @@ const SideBarHeader = () => {
...
   const [results, setResults] = useState<SearchResults | undefined>();
   const isUnmounted = useIsUnmounted();
+
+  // When we add more keyboard shortcuts, we should pull this up and have a CM-like model of the
+  // available actions and their shortcuts, with a hook used here to register a handler for the action.
+  useEffect(() => {
+    const isMac = /Mac/.test(navigator.platform);
+    const keydown = (e: KeyboardEvent) => {
+      if (
+        e.key === "f" &&
```
microbit-robert, line 47:
```suggestion
        (e.key === "F" || e.key === "f") &&
```
Handles default case (capital letter via `shift` key), and lower case if `Caps Lock` is on.

microbit-robert (changes requested):

src/workbench/SideBarHeader.tsx
microbit-matt-hillsdon: I can't even test this as I remapped Caps Lock to Escape long ago!

microbit-robert (approved):
LGTM

PR #474, opened 2022-01-25
Add search test

microbit-robert:
Adds a basic test to open the search modal, input a search term and navigate to the first result.

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -28,6 +28,7 @@ export interface BrowserDownload {
 const defaultWaitForOptions = { timeout: 5_000 };
 
 const baseUrl = "http://localhost:3000";
+const reportsPath = "reports/e2e/";
 
 interface Options {
   /**
@@ -107,13 +108,19 @@ export class App {
       await dialog.accept();
     });
 
-    const logsPath =
-      "reports/screenshots/" + expect.getState().currentTestName + ".txt";
+    const logsPath = reportsPath + expect.getState().currentTestName + ".txt";
+    // Clears previous output from local file.
+    fs.writeFile(logsPath, "", (err) => {
+      if (err) {
+        // Log file error.
+        console.error("Log file error: ", err.message);
+      }
+    });
     page.on("console", (msg) => {
-      fs.appendFile(logsPath, msg.text(), (err) => {
+      fs.appendFile(logsPath, msg.text() + "\n", (err) => {
         if (err) {
           // Log file error.
-          console.log("Log file error: ", err.message);
+          console.error("Log file error: ", err.message);
         }
       });
     });
@@ -764,7 +771,7 @@ export class App {
   async screenshot() {
     const page = await this.page;
     return page.screenshot({
-      path: "reports/screenshots/" + expect.getState().currentTestName + ".png",
+      path: reportsPath + expect.getState().currentTestName + ".png",
     });
   }
 
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -790,6 +790,24 @@ export class App {
     return tab.click();
   }
 
+  async searchToolkits(searchText: string): Promise<void> {
+    const document = await this.document();
+    const searchButton = await document.findByRole("button", {
+      name: "Open search",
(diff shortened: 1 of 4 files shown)


src/e2e/app.ts
```diff
@@ -790,6 +790,24 @@ export class App {
...
+    const searchField = await document.findByRole("textbox", {
+      name: "Search",
+    });
+    await searchField.type(searchText);
+    const modalDialog = await document.findByRole("dialog");
+    // const result = await modalDialog.getAllByRole("button");
+    const result = await modalDialog.findAllByRole("heading", {
+      level: 3,
+    });
+    await result[0].click();
```
microbit-matt-hillsdon, line 808: I think it's not clear from the method name that this clicks on the first result.

Maybe split into two?

microbit-matt-hillsdon (changes requested):
Looks good. One suggestion above.

src/e2e/app.ts
microbit-robert: [`cf28c83`](https://github.com/microbit-foundation/python-editor-next/pull/474/commits/cf28c83a6e090bc5b244e6327e4eade9b7356502)

microbit-matt-hillsdon (approved):

src/e2e/app.ts
```diff
@@ -107,6 +107,17 @@ export class App {
       await dialog.accept();
     });
 
+    const logsPath =
+      "reports/screenshots/" + expect.getState().currentTestName + ".txt";
+    page.on("console", (msg) => {
+      fs.appendFile(logsPath, msg.text(), (err) => {
+        if (err) {
+          // Log file error.
+          console.log("Log file error: ", err.message);
```
microbit-matt-hillsdon, line 116: I think this is fine to be best effort with a logged error, but let's at least do console.error.

src/e2e/app.ts
```diff
@@ -107,6 +107,17 @@ export class App {
       await dialog.accept();
     });
 
+    const logsPath =
+      "reports/screenshots/" + expect.getState().currentTestName + ".txt";
+    page.on("console", (msg) => {
+      fs.appendFile(logsPath, msg.text(), (err) => {
```
microbit-matt-hillsdon, line 113: If you look at the published files they seem to have all the text on one line.
Perhaps we need to add a newline character here?

src/e2e/app.ts
```diff
@@ -107,6 +107,17 @@ export class App {
       await dialog.accept();
     });
 
+    const logsPath =
+      "reports/screenshots/" + expect.getState().currentTestName + ".txt";
```
microbit-matt-hillsdon, line 111: I think it'd now make more sense if "screenshots" was be e.g. "e2e" here and below (ideally using a common constant).

microbit-matt-hillsdon (changes requested):
Some tweaks to the logging needed.

src/e2e/app.ts
microbit-robert: [`109f052`](https://github.com/microbit-foundation/python-editor-next/pull/474/commits/109f052b09bee1277305477bdc62a428562bf793)

src/e2e/app.ts
microbit-robert: [`109f052`](https://github.com/microbit-foundation/python-editor-next/pull/474/commits/109f052b09bee1277305477bdc62a428562bf793)

src/e2e/app.ts
microbit-robert: [`109f052`](https://github.com/microbit-foundation/python-editor-next/pull/474/commits/109f052b09bee1277305477bdc62a428562bf793)

src/e2e/app.ts
microbit-matt-hillsdon: Change looks good. Will just review the output then merge.

microbit-matt-hillsdon:
Looks good now. Interesting to see we don't output screenshots for every case (edits e2e test) but I'll fix that separately.

microbit-matt-hillsdon (approved):

PR #476, opened 2022-01-26
Index dropdown text and labels

microbit-robert:
Indexes dropdown text and labels for explore content.

See #440

### DIFF
--- a/src/documentation/search/search.ts
+++ b/src/documentation/search/search.ts
@@ -127,11 +127,18 @@ const exploreSearchableContent = (toolkit: Toolkit): SearchableContent[] => {
     t.contents?.forEach((e) => {
       const contentString = blocksToText(e.content);
       const detailContentString = blocksToText(e.detailContent);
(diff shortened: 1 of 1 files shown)


src/documentation/search/search.ts
```diff
@@ -127,11 +127,18 @@ const exploreSearchableContent = (toolkit: Toolkit): SearchableContent[] => {
     t.contents?.forEach((e) => {
       const contentString = blocksToText(e.content);
       const detailContentString = blocksToText(e.detailContent);
+      const alternativesLabel = "\n\n" + defaultString(e.alternativesLabel);
```
microbit-matt-hillsdon, line 130: Maybe we should serialize these where they appear in the content?

We could make the blocks -> text conversion support serializers for non-"block" _type blocks. Not sure whether this is worth it but maybe let's discuss later.

microbit-matt-hillsdon: Oh no, ignore me, they're a separate field.

microbit-matt-hillsdon (approved):

PR #477, opened 2022-01-26
Keyboard navigation for search

microbit-robert:
Adds keyboard navigation for search results using tab, as well as up/down arrows. Hover/focus styles added, and viewed search results are tracked and styled appropriately. 

See #440

### DIFF
--- a/src/documentation/search/SearchDialog.tsx
+++ b/src/documentation/search/SearchDialog.tsx
@@ -14,7 +14,7 @@ import {
   Stack,
   Text,
 } from "@chakra-ui/react";
-import { useCallback, useRef } from "react";
+import { useCallback, useEffect, useRef } from "react";
 import { RiCloseLine, RiSearch2Line } from "react-icons/ri";
 import { SearchResults } from "./common";
 import SearchResultList from "./SearchResultList";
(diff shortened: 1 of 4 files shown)


src/documentation/search/SearchDialog.tsx
```diff
@@ -42,6 +46,47 @@ const SearchDialog = ({
...
+        }
+      }
+    };
+
+    document.addEventListener("keydown", keydown);
+    return () => {
+      document.removeEventListener("keydown", keydown);
+    };
+  }, []);
+
```
microbit-robert, line 89: @microbit-matt-hillsdon This is horrible. This overrides the default behaviour of being able to scroll through the search results using the up and down arrows, which on reflection, is not something I'm sure we want to do? If we do want this, perhaps there is a better way?

microbit-matt-hillsdon: Yeah, worth some further thought. Everything else here is more straightforwardly valuable (and fixed tabbing) so perhaps split this bit out and let's get the rest in?

src/documentation/search/SearchResultList.tsx
```diff
@@ -38,37 +48,51 @@ const SearchResultList = ({
...
+    <Stack pl="3px" pr="3px">
+      <Link
+        bgColor={viewedResults.includes(id) ? "#efedf5" : "unset"}
+        borderRadius="md"
+        href={url}
+        onClick={(e) => {
+          e.preventDefault();
+          onClose();
+          setState(navigation);
+          onViewResult(id);
```
microbit-matt-hillsdon, line 76: Maybe the setState call should move up to inside the handler passed for onViewResult?

src/router-hooks.tsx
```diff
@@ -99,14 +110,7 @@ export const RouterProvider = ({ children }: { children: ReactNode }) => {
...
-        .filter(([_, v]) => !!v)
-        .map(([k, v]) => {
-          return `${encodeURIComponent(k)}=${encodeURIComponent(
-            serializeValue(v)
-          )}`;
-        })
-        .join("&");
+      const query = getQueryString(newState);
       const url =
         window.location.toString().split("?")[0] + (query ? "?" + query : "");
```
microbit-matt-hillsdon, line 115: Is this bit common between the new usage and old? Perhaps a `toUrl(state: RouterState)` API would be better and fit both usecases?

microbit-matt-hillsdon (commented):
I really like the highlighting, it feels much more useful than I was expecting.

Added a couple of minor comments.

src/documentation/search/SearchResultList.tsx
microbit-robert: [`f2e178f`](https://github.com/microbit-foundation/python-editor-next/pull/477/commits/f2e178f02edf9d9bf73e5c8a243016d57fb74dba). We now handle closing the on in handleViewResult as well, so this no longer has to be passed down.

src/router-hooks.tsx
microbit-robert: [`1fe7d6e`](https://github.com/microbit-foundation/python-editor-next/pull/477/commits/1fe7d6e76710879f26bd60a083836514a42eb880)

src/documentation/search/SearchResultList.tsx
```diff
@@ -37,38 +44,45 @@ const SearchResultList = ({
...
 
 const SearchResultItem = ({
-  value: { extract, navigation, containerTitle, title },
-  onClose,
+  value: { extract, navigation, containerTitle, title, id },
+  viewedResults,
+  onViewResult,
 }: SearchResultItemProps) => {
-  const [, setState] = useRouterState();
+  const url = toUrl(getQueryString(navigation), navigation);
```
microbit-matt-hillsdon, line 56: Let's move the getQueryString call inside toUrl so the caller just gets a URL from the router state and doesn't have to make two calls.

microbit-robert: [`a905f61`](https://github.com/microbit-foundation/python-editor-next/pull/477/commits/a905f6178a9918ecaed74cf31ef8b850b8005960)

microbit-matt-hillsdon: Change looks great, will just wait for it to build then merge.

microbit-matt-hillsdon (approved):
Looks great.

PR #479, opened 2022-01-26
Try to switch away from homepage to PUBLIC_URL.

microbit-matt-hillsdon:
Set the url to the actual path.

This is motivated by the difficulties on #464 where a worker doing an async import gets the wrong source location as the relative path is resolved relative to the worker script not index.html as it would usually be.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,7 @@ _steps:
   serve: &serve
     run:
       name: Serve files for e2e tests
-      command: npm run serve
+      command: mkdir /tmp/app && cp -r build /tmp/app${PUBLIC_URL} && npx serve --no-clipboard -l 3000 /tmp/app
       background: true
   serve-wait: &serve-wait
     run:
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
This seems to be essentially working now, though the latest e2e run has a novel failure mode:
https://app.circleci.com/pipelines/github/microbit-foundation/python-editor-next/1603/workflows/18719d5e-2f73-4161-a815-99b213091eca/jobs/1677
I can't see how it can be related though. Trying a re-run.

bin/public-url.js
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+const {
+  createDeploymentDetailsWithReviewPrefixes,
+} = require("@microbit-foundation/website-deploy-aws-config");
+
+const {
+  s3Config: { bucketPrefix },
+} = createDeploymentDetailsWithReviewPrefixes();
+console.log(`/${bucketPrefix}/`);
```
microbit-matt-hillsdon, line 16: i.e. the "/update-toolkit-design/" bit in URLs like
https://review-python-editor-next.microbit.org/update-toolkit-design/

.circleci/config.yml
```diff
@@ -99,6 +99,13 @@ jobs:
       - *install_theme
       - *update_version
       - *save_npm_cache
+      # Set PUBLIC_URL based on the bucket prefix. PUBLIC_URL is equivalent to CRA's package.json homepage setting https://create-react-app.dev/docs/advanced-configuration/
+      - run:
+          name: "Set PUBLIC_URL for review branch"
+          environment:
+            NODE_PATH: /usr/local/lib/node_modules
+          # Two env vars as PUBLIC_URL seems to be blank when running jest even if we set it.
+          command: URL=$(npm run --silent public-url); echo "export PUBLIC_URL=$URL && export E2E_PUBLIC_URL=$URL" >> $BASH_ENV
```
microbit-matt-hillsdon, line 108: `$BASH_ENV` is an obscure CircleCI feature. It's set to the path of a file that's sourced as part of the initialisation of the shell used for each step so appending env variable exports to it is a way of ensuring they're available in subsequent steps.

microbit-matt-hillsdon:
Talked this one through with Rob. Merging.

PR #480, opened 2022-01-27
Search translatable strings

microbit-robert:
Add and use search translatable strings.

See #440.


microbit-robert:
What happens to untranslated strings in another language (i.e., French)? FormatJS falls back to using string id with a very loud console error.

microbit-matt-hillsdon (approved):

PR #483, opened 2022-01-27
Adjust reference toolkit spacing

microbit-robert:
Adjust reference toolkit spacing to reduce excess space and ensure highlight on scrolling looks sensible.

### DIFF
--- a/src/documentation/reference/ReferenceNode.tsx
+++ b/src/documentation/reference/ReferenceNode.tsx
@@ -81,14 +81,9 @@ const ReferenceNode = ({
         }}
         fontSize="sm"
         spacing={3}
-        pt={
-          parentType !== "class" ? kindToSpacing[kind] : kindToSpacing[kind] - 1
-        }
-        pb={
-          parentType !== "class"
-            ? kindToSpacing[kind] - 1
-            : kindToSpacing[kind] - 2
-        }
+        // Reduce padding inside a class.
+        pt={kindToSpacing[kind] - (parentType === "class" ? 1 : 0)}
+        pb={kindToSpacing[kind] - (parentType === "class" ? 2 : 1)}
         pl={5}
         pr={3}
         mt={1}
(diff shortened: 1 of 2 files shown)


src/documentation/reference/ReferenceNode.tsx
```diff
@@ -77,9 +77,10 @@ const ReferenceNode = ({ anchor, docs, ...props }: ApiDocEntryNodeProps) => {
         fontSize="sm"
         spacing={3}
         p={5}
+        pb={3}
         pr={3}
         mt={1}
-        mb={kindToSpacing[kind]}
```
microbit-robert, line 82: I think the padding provides plenty of spacing and this only seems to add additional margin where we don't want it? Happy to be proven wrong.

src/documentation/reference/ReferenceNode.tsx
```diff
@@ -77,9 +77,10 @@ const ReferenceNode = ({ anchor, docs, ...props }: ApiDocEntryNodeProps) => {
         fontSize="sm"
         spacing={3}
         p={5}
+        pb={3}
```
microbit-robert, line 80: The reason the reference entries look off is because they start with a code section with no spacing above it, and end with text, or a 'show more' button that both have some inherent spacing below them (the button more than plain text). Reducing the padding here helps counter that, but a bit part of me doesn't like uneven padding.

src/documentation/reference/ReferenceToolkit.tsx
```diff
@@ -74,7 +74,7 @@ const ActiveToolkitLevel = ({
         <List flex="1 1 auto">
           {(module.children ?? []).map((child) => (
             <ListItem key={child.id}>
-              <ReferenceNode docs={child} width="100%" anchor={anchor} mb={0} />
```
microbit-robert, line 77: I'm convinced we don't need `mb={0}` here unless you can find somewhere it is used? I can't see it myself.

src/documentation/reference/ReferenceNode.tsx
microbit-matt-hillsdon: I agree, if anything we still have too much spacing between the methods/fields on a class:

image

PR #484, opened 2022-01-27
Debounce search.

microbit-matt-hillsdon:
This is more necessary when we start looking at wildcard searches or
other kinds that will match more often.


microbit-robert (approved):
Looks good.

PR #486, opened 2022-01-27
Customise the stop word filter.

microbit-matt-hillsdon:
I've hacked it in as the regular list is mostly fine so I don't want to
have to duplicate all its words.

### DIFF
--- a/src/documentation/search/search.test.ts
+++ b/src/documentation/search/search.test.ts
@@ -80,44 +80,10 @@ describe("Search", () => {
     ]);
   });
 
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
Full list is here in case you think I missed anything important: https://github.com/olivernn/lunr.js/blob/aa5a878f62a6bba1e8e5b95714899e17e8150b38/lib/stop_word_filter.js

I've not worried about builtins (e.g. any) that I suspect aren't often used in micro:bit scripts or well documented in our content.

PR #497, opened 2022-01-28
Use drag icon

### LINKED ISSUE #438: [Toolkits] Replace the drag image with an icon
We should try replacing the drag image used when dragging code from Explore or Reference with an icon or some other more concise representation of the code. The code is now shown in the editor as a drop cue.

I don't think we can update the drag image, but if we could then we'd want to update it from the full code to an icon when it's over the code area.

microbit-robert:
Replace default drag image with appropriate icon.

Closes #438.


microbit-robert:
Temporary icon added. Must do better.

microbit-matt-hillsdon:
> Temporary icon added. Must do better.

If I'm reading the offset docs correctly, then I think we need space inside the icon for the cursor, e.g. leave space top left.

We should also:
- share code, e.g. with a common hook useCodeDragImage
- import the image rather than using public/, see e.g. AboutDialog

microbit-matt-hillsdon (approved):
Approving but my changes need review from @microbit-robert

PR #499, opened 2022-01-30
Tweak search extracts to be less scary

### LINKED ISSUE #490: [Search] Improve extracts
Show more context? Show fewer extracts if there are many matches? Don't truncate words? Try to equalise the text length shown?

microbit-matt-hillsdon:
I've gone with showing sentence fragments (based on punctuation) before and after the first position. We then highlight matches in that fragment. Any matches outside it aren't displayed.

This means the amount of text shown is more variable, but the output is much more comprehensible. I think it's generally an improvement.

Closes #490

### DIFF
--- a/src/documentation/search/extracts.test.ts
+++ b/src/documentation/search/extracts.test.ts
@@ -85,7 +85,7 @@ describe("contextExtracts", () => {
       contextExtracts(
         [
           [0, 5], // Extract chosen based on this one
-          [6, 5], // Included due as inside
+          [6, 5], // Included as inside
           [13, 5], // Omitted
         ],
         "match match. match"
--- a/src/documentation/search/extracts.test.ts
+++ b/src/documentation/search/extracts.test.ts
@@ -3,67 +3,128 @@
  *
  * SPDX-License-Identifier: MIT
  */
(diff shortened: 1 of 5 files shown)


src/documentation/search/extracts.test.ts
```diff
@@ -3,67 +3,102 @@
...
+        extract: "match",
+      },
+      {
+        type: "text",
+        extract: " micro:bit",
+      },
+    ]);
+  });
+  it("highlights all matches in returned text", () => {
+    // todo
```
microbit-matt-hillsdon, line 84: Ahem.

src/documentation/search/extracts.test.ts
```diff
@@ -3,67 +3,128 @@
...
   });
-  it("splits text between matches", () => {
+  it("highlights all matches in returned text", () => {
     expect(
       contextExtracts(
         [
-          [0, 6],
-          [16, 6],
+          [0, 5], // Extract chosen based on this one
+          [6, 5], // Included due as inside
```
microbit-matt-hillsdon, line 88:
```suggestion
          [6, 5], // Included as inside
```

microbit-robert (approved):
LGTM. Search results are now much better.

Issue #493, opened 2022-01-27
Review event logging


microbit-matt-hillsdon:
We were quite systematic about logging for the core part of the app, but more recently added features are lacking in this area. Ahead of user testing it would be worth doing a review as it's generally really easy to add.

microbit-matt-hillsdon:
This is now in a good place for user testing. Further work can be tracked separately.

PR #508, opened 2022-01-31
Add navigation logging

microbit-robert:
Add logging for navigation events.

See #493.

### DIFF
--- a/src/router-hooks.tsx
+++ b/src/router-hooks.tsx
@@ -52,9 +52,11 @@ export interface RouterState {
   reference?: Anchor;
 }
 
+type NavigationSource = "toolkit-user" | "toolkit-search" | "toolkit-from-code";
+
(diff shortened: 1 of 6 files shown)


src/router-hooks.tsx
```diff
@@ -110,13 +115,16 @@ export const RouterProvider = ({ children }: { children: ReactNode }) => {
     };
   }, [setState]);
   const navigate = useCallback(
-    (newState: RouterState) => {
+    (newState: RouterState, source?: string) => {
```
microbit-matt-hillsdon, line 118: Please can we type source with e.g. `type NavigationSource = "toolkit-user" | "whatever"` etc.
Otherwise this looks good to me.

microbit-matt-hillsdon (approved):
LGTM, let's merge when CI passes.

Issue #502, opened 2022-01-31
Fix sidebar min width


microbit-matt-hillsdon:
In the absence of work to make it more responsive, we should fix the min-width of the sidebar

![image](https://user-images.githubusercontent.com/44397098/151791571-828f7aa7-6eb8-4be6-a167-000c82ff7e6e.png)

microbit-matt-hillsdon:
We made it respond to the scrollbar width in #561. Closing.

PR #513, opened 2022-02-01
Tweak min width.

microbit-matt-hillsdon:
This needs a more responsive solution, but it needs to be split pane
aware. Going with this for now to avoid broken UI in user testing.

Closes https://github.com/microbit-foundation/python-editor-next/issues/502


microbit-matt-hillsdon:
Closing in favour of https://github.com/microbit-foundation/python-editor-next/pull/561

PR #515, opened 2022-02-01
Add logging for dnd and code insert

microbit-robert:
Add logging for dnd and code insert.

See #493.

### DIFF
--- a/src/editor/active-editor-hooks.tsx
+++ b/src/editor/active-editor-hooks.tsx
@@ -28,10 +28,10 @@ export class EditorActions {
    *
    * @param code The code with any required imports.
    */
-  insertCode = (code: string, parentSlug?: string): void => {
+  insertCode = (code: string, id?: string): void => {
     this.logging.event({
       type: "code-insert",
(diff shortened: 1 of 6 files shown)


src/editor/active-editor-hooks.tsx
```diff
@@ -28,7 +28,11 @@ export class EditorActions {
    *
    * @param code The code with any required imports.
    */
-  insertCode = (code: string): void => {
+  insertCode = (code: string, parentSlug?: string): void => {
```
microbit-matt-hillsdon, line 31: I'd have been tempted to name it something more general by the time it reaches this method, e.g. source or id. If we supported dnd / insert from elsewhere we'd provide some similar identifier but it may not exactly be a "parent slug". 

Similarly for the drag context below.

Where we have existing param documentation it would be good to maintain it.

microbit-matt-hillsdon (changes requested):
One minor thing but no big deal if you feel strongly.

microbit-matt-hillsdon (approved):

PR #517, opened 2022-02-01
Add logging for serial

microbit-robert:
Add logging for serial actions.

See #493.

### DIFF
--- a/src/serial/SerialBar.tsx
+++ b/src/serial/SerialBar.tsx
@@ -1,5 +1,5 @@
 /**
- * (c) 2021, Micro:bit Educational Foundation and contributors
+ * (c) 2022, Micro:bit Educational Foundation and contributors
  *
  * SPDX-License-Identifier: MIT
  */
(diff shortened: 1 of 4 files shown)


src/serial/SerialBar.tsx
```diff
@@ -33,12 +34,25 @@ const SerialBar = ({
...
   const traceback = useDeviceTraceback();
+  const [prevTraceback, setPrevTraceback] = useState<Traceback | undefined>();
+  useEffect(() => {
+    if (traceback && !prevTraceback) {
+      logging.event({
+        type: "serial-traceback",
+      });
+    }
+    setPrevTraceback(traceback);
+  }, [logging, traceback, prevTraceback, setPrevTraceback]);
```
microbit-robert, line 55: This feels clumsy. Is there a better way to get the previous traceback? We could change the hook, but it seems like more effort than it's worth when we only need the previous traceback here.

microbit-robert: Improved after discussion with @microbit-matt-hillsdon. See [`fa77951`](https://github.com/microbit-foundation/python-editor-next/pull/517/commits/fa779517cc9338cdec9e6094cc9d18589d94e0b4).

microbit-matt-hillsdon (approved):

PR #518, opened 2022-02-01
Code embed: remove pop-up code in favour of z-index

microbit-matt-hillsdon:
This needs more testing before I'm confident it's the right direction, in particular I've focussed at 

I also clear the pop-up on scroll which seems to fix the sticky pop-ups on scroll issue.

You still can't scroll when dwelling over a pop-up. I'm not really sure why. It seems like this might be unavoidable.

### DIFF
--- a/src/documentation/explore/CodeEmbed.tsx
+++ b/src/documentation/explore/CodeEmbed.tsx
@@ -5,25 +5,14 @@
  */
 import { Button } from "@chakra-ui/button";
 import { Box, BoxProps, HStack } from "@chakra-ui/layout";
-import { Portal } from "@chakra-ui/portal";
 import { forwardRef } from "@chakra-ui/system";
-import {
-  Ref,
-  RefObject,
-  useCallback,
-  useLayoutEffect,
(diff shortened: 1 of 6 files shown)


microbit-matt-hillsdon:
This is noticeably worse in Safari but seems OK in FireFox and Chrome.

The scrollbar overlaps the pop-up (except occasionally it doesn't) and if you drag to highlight text then you can scroll the sidebar to the right even though it has no scrollbars.

I'd consider pointerEvents="none" for Safari (especially if we added a copy button) but I think we'd need a workaround for the scrollbar issue at a minimum.

Even if we can't progress with this, there might be ideas we can take from it to help improve the pop-up approach.

microbit-matt-hillsdon:
If we do go with this then we need to reinstate the shadow and review the styling for other omissions. Not worth looking at unless we find a Safari fix.

microbit-matt-hillsdon:
Discarding this approach because of the Safari difficulties, but it'd be nice to revisit in future if they fix it.

PR #519, opened 2022-02-01
Log search

microbit-robert:
Add logging when a the search query changes from undefined and a search is performed.

See #493.

### DIFF
--- a/src/workbench/SideBarHeader.tsx
+++ b/src/workbench/SideBarHeader.tsx
@@ -26,6 +26,7 @@ import { topBarHeight } from "../deployment/misc";
 import { SearchResults } from "../documentation/search/common";
 import { useSearch } from "../documentation/search/search-hooks";
 import SearchDialog from "../documentation/search/SearchDialog";
+import { useLogging } from "../logging/logging-hooks";
 import { RouterState, useRouterState } from "../router-hooks";
 
 const SideBarHeader = () => {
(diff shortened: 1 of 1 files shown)


src/workbench/SideBarHeader.tsx
```diff
@@ -66,16 +68,19 @@ const SideBarHeader = () => {
...
+            }
+            setResults(newResults);
           }
         } else {
           setResults(undefined);
         }
         setViewedResults([]);
       }, 300),
-    [search, setResults, setViewedResults, isUnmounted]
+    [search, setResults, setViewedResults, isUnmounted, logging, results]
```
microbit-matt-hillsdon, line 83: Can we use the same trick as we discussed on zoom to avoid depending on results here by passing a function to setResults? Nothing else we depend on is expected to change.

microbit-matt-hillsdon (changes requested):
One suggestion, otherwise looks good.

src/workbench/SideBarHeader.tsx
microbit-robert: Please see [`a2b0e40`](https://github.com/microbit-foundation/python-editor-next/pull/519/commits/a2b0e40ef71c0079052dc2d5e10cdc66ff3c57e5).

microbit-matt-hillsdon (approved):

PR #521, opened 2022-02-02
Logging fixes

microbit-robert:
Change some logging events to use `message` instead of `detail` for analytics. 

Move logging event for `create-file` so it only fires when a file is created, not when the dialog is opened.

See #493.


microbit-matt-hillsdon (approved):
Looks good to me.

PR #529, opened 2022-02-02
Snap sidebar open/closed

microbit-robert:
Add ability to close and open sidebar by dragging split view handle beyond min limits.


microbit-matt-hillsdon:
We went with another version of this. Closing.

PR #530, opened 2022-02-02
Allow digits in filenames.

### LINKED ISSUE #524: "Create file" confusingly rejects filenames with digits/numbers/numerals
image

It doesn't mention that the digits are a problem. We should check the real restriction and either adjust it or the error text.

Reported by @microbit-katie

microbit-matt-hillsdon:
Closes #524.

### DIFF
--- a/src/project/project-utils.test.ts
+++ b/src/project/project-utils.test.ts
@@ -34,5 +34,6 @@ describe("validateNewFilename", () => {
     expect(
       validateNewFilename("underscores_are_allowed", exists, intl)
     ).toBeUndefined();
+    expect(validateNewFilename("007digits", exists, intl)).toBeUndefined();
   });
 });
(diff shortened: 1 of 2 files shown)


microbit-robert (changes requested):
While you can have python files/modules that start with a number run okay in isolation, Python doesn't let you import from files/modules that start with a number. We either need to disallow numbers again with the appropriate error message, or cater for Python's naming conventions by changing the regex (and probably adding another message to the user).

microbit-matt-hillsdon:
Thanks, obvious now you point it out! 
Makes me less inclined to do this partial fix.
The original was based on https://www.python.org/dev/peps/pep-0008/#package-and-module-names (which I failed to find yesterday). Will revisit this later.

microbit-matt-hillsdon:
Support ticket: https://support.microbit.org/a/tickets/54238 

This was just raised independently via support.

microbit-matt-hillsdon:
Superseded by https://github.com/microbit-foundation/python-editor-next/pull/770

PR #532, opened 2022-02-03
Fix unintended change to stacking contexts.

microbit-matt-hillsdon:


src/documentation/explore/Slide.tsx
```diff
@@ -48,10 +58,12 @@ const Slide = ({
   children: JSX.Element;
 }) => {
   const reducedMotion = useReducedMotion();
+  const ref = useRef<HTMLDivElement>(null);
   return reducedMotion ? (
     children
   ) : (
     <motion.div
+      ref={ref}
```
microbit-matt-hillsdon, line 66:
```suggestion
  const reducedMotion = useReducedMotion();
  return reducedMotion ? (
    children
  ) : (
    <motion.div
```

microbit-matt-hillsdon:
@microbit-robert, I don't think this changes anything but lets get it in anyway to prevent future confusion.

microbit-robert (approved):

PR #542, opened 2022-02-03
Remove redundant aria-label.

microbit-matt-hillsdon:
The distinction between this and search is also lost in translation so
another reason to lose it.


microbit-robert (approved):
LGTM.

Issue #450, opened 2022-01-20
Make “my program” a translatable string?


microbit-matt-hillsdon:
One for discussion. It's currently stored along with the initial program which isn't localised.

![image](https://user-images.githubusercontent.com/44397098/152179907-71cd5040-c8f8-4fa3-b80d-961f23bde222.png)

microbit-matt-hillsdon:
I think I didn't translate this as it wasn't translated in V2, but I think it's worth discussion:

https://python.microbit.org/v/2?lang=fr

microbit-matt-hillsdon:
We should localise "my program" and the comment at the top of the script.

If we can let's do it for user testing.

microbit-matt-hillsdon:
Let's discuss the comment here (and in a new file) further. Is there any value in either of them?

microbit-matt-hillsdon:
Short term: change to "Untitled project", make translatable.
Ensure that changing language before user edit changes text but not after.
Do this for user testing.

We don't think the comment at the start has much value. We'd like to encourage edits and reference Explore.
We should make it translatable.
@microbit-mark to suggest text for discussion.

microbit-mark:
Some permutations in https://paper.dropbox.com/doc/Python-script-intro--BbPcjaFoaBrHPGGzuz97CepcAg-sOSEZvjTdGZRytYWq330V

I would suggest something that is explicit and related to the most common user action of editing the code/script and flashing it to the device

![image](https://user-images.githubusercontent.com/31242877/152361418-4f73a9bd-c260-4019-9f69-433f1236e502.png)

I have tried some versions where we reference the Explore menu, but I wonder if doing so might make it really confusing if you take the script out of context. Eg if I open it in a text editor and it's talking about an 'Explore' menu

![image](https://user-images.githubusercontent.com/31242877/152361713-10cf093a-07d6-428b-bbe2-0b5fed00c581.png)

microbit-matt-hillsdon:
We should consider some kind of non-code notice/nudge too, if the main point is to communicate the above rather than to demonstrate a code comment.

microbit-matt-hillsdon:
Moved the issue of the default script to #566 along with @microbit-mark's suggestions.

Closing this as we've addressed the project name.

PR #545, opened 2022-02-03
Translate project name

microbit-matt-hillsdon:
Don't set a project name in the fs unless the user does, so we can
default it in the UI layer based on the current language.

See #450

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -272,12 +272,12 @@ export class App {
     await button.click();
   }
 
-  async openSettingsDialog(): Promise<void> {
-    return this.findAndClickButton("Settings");
-  }
-
-  async closeSettingsDialog(): Promise<void> {
-    return this.findAndClickButton("Close");
+  async switchLanguage(locale: string): Promise<void> {
+    // All tests ids so they can be language invariant.
+    const document = await this.document();
+    await (await document.findByTestId("settings")).click();
+    await (await document.findByTestId("language")).click();
+    await (await document.findByTestId(locale)).click();
   }
 
   /**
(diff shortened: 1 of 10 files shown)


src/e2e/app.ts
```diff
@@ -272,12 +272,12 @@ export class App {
...
   }
 
-  async openSettingsDialog(): Promise<void> {
-    return this.findAndClickButton("Settings");
-  }
-
-  async closeSettingsDialog(): Promise<void> {
-    return this.findAndClickButton("Close");
+  async switchLanguage(locale: string): Promise<void> {
+    // All tests ids so they can be language invariant.
```
microbit-matt-hillsdon, line 276:
```suggestion
    // All test ids so they can be language invariant.
```

src/e2e/settings.test.ts
```diff
@@ -11,8 +11,11 @@ describe("Browser - settings", () => {
   afterEach(app.screenshot.bind(app));
   afterAll(app.dispose.bind(app));
 
-  it("shows the settings dialog when the button is clicked", async () => {
```
microbit-matt-hillsdon, line 14: This pre-existing test was broken but we didn't notice as the file name didn't match the pattern the test runner uses. It didn't do anything worthwhile so I've just removed it. It dated to when settings was a top-level menu.

microbit-robert (approved):
LGTM.

microbit-matt-hillsdon:
Spotted an issue where the project name doesn't seem to make it to session storage, so doesn't survive a refresh or tab close/reopen. It's not new though; it affects main too.  Raised #547.

PR #546, opened 2022-02-04
Improve navigation highlight

microbit-robert:
This prevents a gap between the Highlight element and the scroll bar (see below) for class methods/fields in the reference documentation.

image


microbit-matt-hillsdon (approved):

microbit-matt-hillsdon:
Looks much better, thanks.

Issue #509, opened 2022-02-01
Use a third-party routing solution


microbit-matt-hillsdon:
We have some homebrew URL management, started when it seemed our approach to URLs might be a bad fit for react/reach-router. There's no reason to believe that now. Assuming we continue to think links/URL changes are the best way to navigate the toolkit content then we should switch to something third-party.

PR #552, opened 2022-02-04
Fix search results navigation

### LINKED ISSUE #535: Clicking same search result twice has no effect
On clicking a search result, navigation to the clicked topic entry is activated. Clicking the same search again (without clicking any other result, or navigating away from the open topic) has no effect. This is problematic if the user has scrolled through the topic, and wishes to return the searched entry using the results.

microbit-robert:
Clicking same entry twice in a row (or more) now results in the correct navigation behaviour.

Closes #535.


microbit-matt-hillsdon (approved):
I worry about the non-obviousness of this mechanism but let's consider that on #509.

Issue #534, opened 2022-02-03
Code popup hidden by breadcrumb in some situations


microbit-robert:
After using search to navigate from one topic to another, the code popups for that topic appear behind the breadcrumb, but above other elements. This might be related to transform properties leftover from slide transitions.

![image](https://user-images.githubusercontent.com/95928279/152356560-011552e8-8430-4f7f-b2ca-57a85789ae94.png)

microbit-matt-hillsdon:
@microbit-robert, can you still reproduce this?

microbit-matt-hillsdon:
@microbit-robert, sorry, can repro now!

microbit-matt-hillsdon:
Closed by #554.

PR #554, opened 2022-02-04
Put code pop-up above header.

microbit-matt-hillsdon:
@microbit-robert I could only get the behaviour in the screenshot on #534 on main.

I think it's better flipped, which I've done on this PR.

Can you reproduce varying behaviour anymore on main or this PR?


microbit-robert (approved):
Looks good. Can't reproduce the issue on this branch.

PR #561, opened 2022-02-04
Collapse the search button to an icon when the split view is smaller

microbit-matt-hillsdon:


src/common/CollapsibleButton.tsx
```diff
@@ -33,9 +38,16 @@ export type CollapsableButtonComposibleProps = Omit<
  *
  * We'd like to do this at a lower-level so we can animate a transition.
  */
-const CollapsableButton = React.forwardRef(
+const CollapsibleButton = React.forwardRef(
```
microbit-matt-hillsdon, line 41: So it turns out I can't spell 😣

microbit-matt-hillsdon:
In case it seems a little off, the point of collapse aims to fit the French text as well as the English.

microbit-robert (approved):
Looks good.

Issue #547, opened 2022-02-04
Project name not retained after refresh


microbit-matt-hillsdon:
We use session storage to store the code. It was my intention to do likewise with the project name.
While testing #545 I spotted that this doesn't work (the PR does not affect the behaviour). A refresh after changing the project name and the code only retains the code change.

PR #562, opened 2022-02-07
Fix session storage for project name.

microbit-matt-hillsdon:
We were storing it just fine, but not initializing with it.

Closes https://github.com/microbit-foundation/python-editor-next/issues/547

### DIFF
--- a/src/fs/storage.ts
+++ b/src/fs/storage.ts
@@ -236,10 +236,10 @@ const initializeFromStorage = async (from: FSStorage, to: FSStorage) => {
   if (projectName) {
     await to.setProjectName(projectName);
   }
-  return Promise.all([
-    ...files.map(async (f) => {
+  return Promise.all(
+    files.map(async (f) => {
       const v = await from.read(f);
       return to.write(f, v);
-    }),
-  ]);
+    })
+  );
 };
--- a/src/fs/storage.ts
+++ b/src/fs/storage.ts
@@ -19,7 +19,7 @@ export interface FSStorage {
   read(filename: string): Promise<Uint8Array>;
   write(filename: string, content: Uint8Array): Promise<void>;
   remove(filename: string): Promise<void>;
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
> Hmm, looks like e2e failure could be a real issue. Looking now.

Fixed in [db6a419](https://github.com/microbit-foundation/python-editor-next/pull/562/commits/db6a4198a1716e7d597e199c72ab0c405f73a4b3)

microbit-robert (approved):
LGTM.

src/fs/storage.ts
```diff
@@ -93,11 +94,15 @@ export class SessionStorageFSStorage implements FSStorage {
   }
 
   async setProjectName(projectName: string) {
-    this.storage.setItem("projectName", projectName);
+    if (projectName === undefined) {
+      this.storage.removeItem(projectNameKey);
+    } else {
+      this.storage.setItem(projectNameKey, projectName);
+    }
```
microbit-matt-hillsdon, line 101: Oops, I'd changed the signature to support undefined at some point, reverted it, and left this nonsense in.

```suggestion
    this.storage.setItem(projectNameKey, projectName);
```

src/fs/storage.ts
```diff
@@ -225,12 +226,16 @@ export class SplitStrategyStorage implements FSStorage {
...
+const initializeFromStorage = async (from: FSStorage, to: FSStorage) => {
   const files = await from.ls();
-  return Promise.all(
-    files.map(async (f) => {
+  const projectName = await from.projectName();
+  if (projectName) {
+    await to.setProjectName(projectName);
+  }
+  return Promise.all([
+    ...files.map(async (f) => {
```
microbit-matt-hillsdon, line 236: Oops, this is also odd, from when I updated project name unconditionally and so put it inside the Promise.all. The extra list is now pointless.

PR #568, opened 2022-02-08
CRA upgrade

microbit-robert:
Upgrade to create react app version 5.

PR #569, opened 2022-02-09
Craco upgrade

microbit-robert:
Upgrade craco to alpha version to support CRA v5.


microbit-matt-hillsdon:
Rolling this in with #568. Closing separate PR.

PR #574, opened 2022-02-09
Search refactor

microbit-robert:
Move search logic into `search-hooks.tsx` and clear search query and results when language is changed.

### DIFF
--- a/src/documentation/search/search-hooks.tsx
+++ b/src/documentation/search/search-hooks.tsx
@@ -3,7 +3,6 @@
  *
  * SPDX-License-Identifier: MIT
  */
-import { DebouncedFunc } from "lodash";
 import debounce from "lodash.debounce";
 import {
   createContext,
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
This looks good to me now. @microbit-robert, feel free to merge it once you've taken a look at my changes.

microbit-robert:
@microbit-matt-hillsdon There is now a slight delay in closing the search results on clearing the search input, because we no longer directly set results as undefined, and we let the de-bounced search do it for us.

src/documentation/search/search-hooks.tsx
```diff
@@ -22,15 +39,53 @@ export const useSearch = (): Search => {
...
+    [setResults, isUnmounted, logging]
+  );
+
+  const [{ languageId }] = useSettings();
+  useEffect(() => {
+    setQuery("");
+  }, [languageId]);
+
+  useEffect(() => {
+    debouncedSearch(query);
```
microbit-robert, line 80: Something like this prevents the delay in closing the results on clearing the search input.
```suggestion
    if (!query.trim()) {
      setResults(undefined);
    } else {
      debouncedSearch(query);
    }
```

microbit-robert: [`[6d15abb](https://github.com/microbit-foundation/python-editor-next/pull/574/commits/6d15abb74b7269c1afd42518e6c0246210169ba1)`](https://github.com/microbit-foundation/python-editor-next/pull/574/commits/6d15abb74b7269c1afd42518e6c0246210169ba1)

microbit-matt-hillsdon (approved):

PR #579, opened 2022-02-10
Fix traceback name and link for v1

### LINKED ISSUE #571: Navigation from serial to program doesn't work for main file with v1 MicroPython
Use a V1 and create a traceback. Note the strange __main__ which isn't normally shown:

image

Navigation to the line doesn't work. It seems to scroll to the top.

microbit-robert:
Fixes the traceback name and link for v1 boards where the error occurs in the `main` file.

Closes #571.

### DIFF
--- a/src/device/device-hooks.test.tsx
+++ b/src/device/device-hooks.test.tsx
@@ -136,4 +136,23 @@ SomeError:
     traceback = tsb.push(toCrLf("0".repeat(5000)))!;
     expect(traceback).toBeUndefined();
   });
+
+  it("provides the correct main file name for V1 and V2 boards", () => {
+    const tsb = new TracebackScrollback();
+    // Entry/main file is __main__ for V1 boards
+    const tracebackV1 = tsb.push(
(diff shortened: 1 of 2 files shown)


microbit-robert:
Should we also prevent users from creating a file called `__main__`. In this example you can import from `__test__` but you receive an import error importing from `__main__` (as expected really).

image

microbit-matt-hillsdon (changes requested):
Please would you add a test with V1 output that clearly documents the two cases as being V1 and V2 so we don't accidentally remove V1 support in future? This code is fairly well tested in device-hooks.test.ts so should be easy to add.

Change itself looks good.

microbit-matt-hillsdon:
> Should we also prevent users from creating a file called `__main__`. In this example you can import from `__test__` but you receive an import error importing from `__main__` (as expected really).

I think this is unlikely in practice and probably not worth spending time on.

microbit-matt-hillsdon (approved):
Looks good.

Issue #449, opened 2022-01-20
Investigate text drag-and-drop


microbit-matt-hillsdon:
Something is wrong with text drag and drop, perhaps due to our own drag and drop handling.

If you drag code you can make it disappear.

Highlight like this and drag from the indent area. Sometimes it breaks, sometimes it doesn't drag at all.

![image](https://user-images.githubusercontent.com/44397098/150374920-04a47316-0eb2-4dc2-9b7e-2952d4b8034a.png)

microbit-robert:
#581 fixes the issue of text disappearing on drag, however, no cursor is shown when dragging text inside the editor, so it is hard to pin point exactly where the text will drop. `@codemirror/view@0.19.31` introduces the `dropCursor` extension which fixes this issue. We should add this to the CodeMirror config when we have upgrade CodeMirror to the latest version.

PR #581, opened 2022-02-11
Fix text drag inside editor

microbit-robert:
Only prevent default behaviour when the drag event contains files so that default CodeMirror behaviour is preserved when dragging and dropping text into and within the editor.

See #449.


microbit-matt-hillsdon (approved):
Looks good.

PR #591, opened 2022-02-14
Multiple sidebar changes

### LINKED ISSUE #584: Move Reference down with Files, change icon to multiple files not folder, add projects placeholder
To address the concern that folks end up in Reference too easily, when they'd be better served by the content in Explore, we should:

- [ ] Move Reference down with Files
- [ ] Change icon to multiple files not folder. There isn't one in remix, try VscFiles ?
- [ ] Add projects placeholder below Explore. Unsure about this. What icon?

microbit-robert:
Moves Reference down with Files and mutes colour. 
Changes Files icon to multiple files.
Adds Projects placeholder below Explore.

Closes #584.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -383,6 +383,14 @@
     "defaultMessage": "Project header",
     "description": "Aria label for the project header area"
   },
+  "projects-tab": {
+    "defaultMessage": "Projects",
+    "description": "Projects tab button text"
+  },
+  "projects-tab-description": {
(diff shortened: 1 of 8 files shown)


microbit-matt-hillsdon:
> "Move Reference down with Files and mutes colour."

Let's try reverting this part, on the basis that the additional of Projects helps push Reference out the way a bit already, and Explore/Projects/Reference will all be in scope for search.

microbit-matt-hillsdon (approved):

PR #595, opened 2022-02-14
Update typeshed.json file

microbit-robert:
Update typeshed.json file following change to stubs.


microbit-matt-hillsdon:
We should confirm that mro is gone from e.g. `Image.`'s autocomplete.

microbit-matt-hillsdon (approved):
LGTM

PR #596, opened 2022-02-14
Translated via Crowdin.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
Not worth review for the loss of a 't' 🙂 
Merging.

Issue #592, opened 2022-02-14
Broken cursor position when accepting completion directly after dot


microbit-matt-hillsdon:
![image](https://user-images.githubusercontent.com/44397098/153887026-334e2776-c2ae-4f8b-b6e0-410f33725336.png)

I think I accepted the completion with Enter and then immediately typed a comma.

microbit-matt-hillsdon:
The problem is that selecting an option without typing any text (as is likely with Image or Sound) doesn't move the cursor/selection.

microbit-matt-hillsdon:
The same is true if you trigger completion with no text and select something.
No similar issue exists on the CM demo on their website so likely our fault.

microbit-matt-hillsdon:
I don't really understand the inconsistency but we're doing no selection handling at all when we override apply in the completion code. CM's default implementation is:

```
    view.dispatch({
      changes: {from: result.from, to: result.to, insert: apply},
      selection: {anchor: result.from + apply.length},
      userEvent: "input.complete",
      annotations: pickedCompletion.of(option.completion)
    })
```

ours should be broadly similar, after taking care of the bracket insertion issue.

PR #597, opened 2022-02-14
Fix autocomplete selection with no brackets.

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-next/issues/592


microbit-robert (approved):
Looks good.

PR #610, opened 2022-02-21
Avoid stuck sidebar animation

### LINKED ISSUE #573: Animation stuck after double click on sidebar tab
1. Navigate to an Explore topic.
1. Double click on the Explore tab.
1. Observe stuck animation.

Fixable by navigating to a topic or Reference but not by clicking again on Explore

image

microbit-robert:
Avoids stuck sidebar animation when double clicking on a sidebar tab to go back to the top level.

Closes #573.

### DIFF
--- a/src/workbench/SideBar.tsx
+++ b/src/workbench/SideBar.tsx
@@ -100,7 +100,7 @@ const SideBar = ({
     ];
     return result;
   }, [onSelectedFileChanged, selectedFile, intl]);
-  const [{ tab }, setParams] = useRouterState();
+  const [{ tab, explore, reference }, setParams] = useRouterState();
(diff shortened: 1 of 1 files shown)


src/workbench/SideBar.tsx
```diff
@@ -115,10 +115,12 @@ const SideBar = ({
     // A click on a tab when it's already selected should
     // reset any other parameters so we go back to the top
     // level.
-    setParams({
-      tab,
-    });
-  }, [tab, setParams]);
+    if (explore || reference) {
```
microbit-matt-hillsdon, line 118: This is quite subtle. I'm ok with merging but wonder whether it'd be possible to fix closer to the animation.

PR #613, opened 2022-02-22
Add dropCursor to CodeMirror config

### LINKED ISSUE #449: Investigate text drag-and-drop

microbit-robert:
Adds `dropCursor` to CodeMirror config which draws a cursor at the current drop position when something is dragged over the editor.

Closes #449.

### DIFF
--- a/src/editor/codemirror/dropcursor.ts
+++ b/src/editor/codemirror/dropcursor.ts
@@ -91,7 +91,7 @@ const drawDropCursor = ViewPlugin.fromClass(
           this.cursor!.className = "cm-dropCursor";
         }
         if (
-          update.startState.field(dropCursorPos) != cursorPos ||
+          update.startState.field(dropCursorPos) !== cursorPos ||
           update.docChanged ||
           update.geometryChanged
         )
@@ -128,7 +128,7 @@ const drawDropCursor = ViewPlugin.fromClass(
     }
 
     setDropPos(pos: number | null) {
-      if (this.view.state.field(dropCursorPos) != pos)
+      if (this.view.state.field(dropCursorPos) !== pos)
         this.view.dispatch({ effects: setDropCursorPos.of(pos) });
     }
   },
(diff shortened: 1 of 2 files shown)


microbit-robert:
@microbit-matt-hillsdon This works well in general, but might be confusing for users using our dnd implementation where we change the selection after the drop.

microbit-matt-hillsdon:
Maybe we just copy https://github.com/codemirror/view/blob/main/src/dropcursor.ts with attribution and license at the top as well as a link to the original source then make a small tweak to filter by content type so it's disabled when dropping code snippets?

We could request a feature in CM but I'm not sure it's reasonable one.

src/editor/codemirror/dropcursor.ts
```diff
@@ -0,0 +1,166 @@
...
+  },
+  {
+    eventHandlers: {
+      dragover(event) {
+        // Avoid setting drop position if dragged element is a code snippet/example.
+        if (!(event.target as HTMLElement).classList.contains("cm-content")) {
+          this.setDropPos(
+            this.view.posAtCoords({ x: event.clientX, y: event.clientY })
+          );
+        }
```
microbit-robert, line 143: @microbit-matt-hillsdon This works but feels like it could be flimsy. Open to other suggestions to determine what is being dragged.

microbit-matt-hillsdon: One of event.dataTransfer.types is equal to our `pythonSnippetMediaType` constant?

microbit-robert: Perfect - [`94eaf98`](https://github.com/microbit-foundation/python-editor-next/pull/613/commits/94eaf98b0bd76e07c988cbc40e97f279fc39db09)

microbit-matt-hillsdon (approved):
Noted that on Mac Chrome you have to wait a bit before drag. This seems to be a difference in behaviour between platforms and is unrelated to the changes we've made.

Works well now.

Issue #602, opened 2022-02-17
Translation bug in serial interface


microbit-mark:
Identified in user testing.

The French translation for show/hide serial refers to the "serial number" not the console. 

![image](https://user-images.githubusercontent.com/31242877/154502822-a5530cba-e051-4b00-a6b4-fffa6b1a53bc.png)

microbit-matt-hillsdon:
I'd like to get this fixed next week, before Mark does further user testing which we're hoping will have more French users.

microbit-matt-hillsdon:
I confirmed that the context updates cleanly in Crowdin.

PR #614, opened 2022-02-22
Clearer Crowdin context.

microbit-matt-hillsdon:
See https://github.com/microbit-foundation/python-editor-next/issues/602


microbit-robert (approved):
LGTM.

Issue #605, opened 2022-02-17
Disable autocorrect for file creation dialog


microbit-matt-hillsdon:
On Safari, type lowercase "dave" in the file creation box. Tab out of the field, it's "corrected" to "Dave" which results in a validation error:

![image](https://user-images.githubusercontent.com/44397098/154561602-589bbad9-a496-43ba-acda-db614d57d4e4.png)

PR #621, opened 2022-02-24
Disable auto fix-ups for the new file input dialog.

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-next/issues/605

None of the automatic behaviours seem useful for file names that may well be abbreviations/contractions/technical terms etc.


microbit-matt-hillsdon:
I've confirmed this resolves the reported issue in Safari.

microbit-robert (approved):
LGTM.

Issue #236, opened 2021-08-11
Preparatory work to integrate with micro:bit classroom


microbit-matt-hillsdon:
## Description

The V2 editor supports an iframe-based embedding protocol that allows it to be embedded in micro:bit classroom and similar environments.

The new editor should support the same protocol.

The V2 editor protocol just uses the text of the main.py script as the project (unlike MakeCode, which uses the MakeCode JSON project structure). We should support a richer project structure for Python projects with the new editor supporting multiple files. We should discuss whether this should be JSON as per MakeCode or something based on the binary file system format.

We'd like to offer an option to opt in to using the beta Python editor with micro:bit classroom rather than just cut over so this ticket also covers a UI.

## Todo
- [x] Proof of concept with main.py only
- [x] Fix initialisation so that we don't have a main.py until it has the content from the host

- [x] Review protocol, agree project format (2d, more elapsed).
- [x] Implement in editor (1d).
- [x] Classroom: Add URL parameter or similar to allow testing with the alpha (0.5).
- [x] Classroom UI: Add a file drop-down to the Python rendering on the Student code page? (0.5)
    - This bit wasn't discussed.
- [x] Classroom UI: List all non-tagged module files in the Word doc, main.py first then in alphabetical order. (0.5)
    - This bit wasn't discussed. Makes sense to exclude support code?
- [x] Ticket in micro:bit classroom with proposed UI change to opt in to beta use (expect ~0.5 to implement assuming we do a select on the same row as Python)

Potential further work to improve classroom to cope with messages over Pusher that exceed our limit (20k compressed?), if we think we're more likely to hit the limit with this version of the editor.

## Conclusion

...

microbit-matt-hillsdon:
This just needs some UI review, perhaps in the Python meeting today. @microbit-robert please can you be in a position to demo with reasonably quick access to the new bits of UI.

microbit-matt-hillsdon:
We tweaked the UI after discussion to use an additional radio group option.

Closing.

PR #623, opened 2022-02-28
Handle initial project for embedded editor

microbit-robert:
Initial project code now handles case where editor is embedded. File system is not initialised until content from host is received in this case.

See #236.


microbit-robert:
@microbit-matt-hillsdon I've removed a couple of unused imports in `host.test.ts`. This looks good to me.

src/fs/host.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+    logging.error(
+      "Cannot detect valid host controller despite controller URL parameter."
+    );
+  }
+};
+
+const setMainCode = (fs: FileSystem, code: string): void => {
+  fs.write(MAIN_FILE, code, VersionAction.INCREMENT);
+};
```
microbit-matt-hillsdon, line 132: This is now a little weird as it's not reused, might as well inline it.

microbit-matt-hillsdon (approved):
Looks good now.

Issue #478, opened 2022-01-26
`microbit` is not available as an import option to autocomplete


microbit-mark:
## Bug Description

Typing `from microbit import *` I would expect the `microbit` module to be available to import

## Screenshots

![image](https://user-images.githubusercontent.com/31242877/151194051-36f4fa84-d3c2-4f65-84a6-f1729192226e.png)

Chrome 97, MacOS 12

microbit-matt-hillsdon:
This is fixed by #624 due to changes in Pyright. That explains why it was already working in VS Code (we've just merged in some three months of upstream work).

Closing.

microbit-matt-hillsdon:
Note this is just building now on alpha: https://app.circleci.com/pipelines/github/microbit-foundation/python-editor-next/1880/workflows/ae50431f-c37f-4209-8187-656fc56a01ff

PR #624, opened 2022-03-01
Update pyright

### LINKED ISSUE #572: Update Pyright
1. Merge latest changes from pyright into browser branch of our fork and confirm it works.
2. Merge browser branch into microbit branch.
3. Update in app and test cross browser.

Is it significantly larger? We might want to look at disabling features we don't need on the micro:bit branch if there are clean(ish) seams we can use.

microbit-robert:
Use latest Pyright build after updating Pyright from upstream.

Closes #572.


microbit-matt-hillsdon:
This fixes #478, which makes sense as it was already working for me in VS Code.

There's a 155k file size increase (before compression). I think we should still merge and plan work to address. There's likely code from our webpack we can drop if we put the time in to analyse the codebase more (not necessarily what was added this time).

microbit-matt-hillsdon (approved):
This is looking good to me. Reference seems to be unchanged.

PR #626, opened 2022-03-02
Handle embedded editor with multi-file python projects

microbit-robert:
Add support to communicate python projects with multiple files with host.

See #236.

### DIFF
--- a/src/fs/fs.ts
+++ b/src/fs/fs.ts
@@ -352,8 +352,7 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
       files: {},
       projectName,
     };
-    const fs = await this.initialize();
-    for (const file of await fs.ls()) {
+    for (const file of await this.storage.ls()) {
       const data = await this.storage.read(file);
       const content = new TextDecoder().decode(data);
       project.files[file] = content;
@@ -364,10 +363,11 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
   async replaceWithMultipleFiles(
     project: MultiFilePythonProject
   ): Promise<void> {
-    const fs = await this.initialize();
-    fs.ls().forEach((f) => fs.remove(f));
+    for (const file of await this.storage.ls()) {
+      await this.remove(file);
+    }
     for (const key in project.files) {
-      fs.write(key, project.files[key]);
+      this.write(key, project.files[key], VersionAction.INCREMENT);
     }
     this._dirty = false;
     this.project = {
--- a/src/fs/fs.ts
+++ b/src/fs/fs.ts
@@ -21,12 +21,7 @@ import {
   SessionStorageFSStorage,
   SplitStrategyStorage,
 } from "./storage";
-import {
-  InitialProject,
-  MultiFilePythonProject,
-  isSimplePythonProject,
-  SimplePythonProject,
-} from "./initial-project";
+import { PythonProject } from "./initial-project";
 
 const commonFsSize = 20 * 1024;
 
@@ -208,7 +203,7 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
    * We remember this so we can tell whether the user has edited
    * the project since for stats generation.
    */
-  private cachedInitialProject: InitialProject | undefined;
+  private cachedInitialProject: PythonProject | undefined;
 
   async initialize(): Promise<MicropythonFsHex> {
     if (this.fs) {
@@ -219,10 +214,15 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
         if (!(await this.exists(MAIN_FILE))) {
           // Do this ASAP to unblock the editor.
           this.cachedInitialProject = await this.host.createInitialProject();
-          if (isSimplePythonProject(this.cachedInitialProject)) {
-            this.initializeSimplePythonProject(this.cachedInitialProject);
-          } else {
-            this.initializeMultiFilePythonProject(this.cachedInitialProject);
+          if (this.cachedInitialProject.projectName) {
+            await this.setProjectName(this.cachedInitialProject.projectName);
+          }
+          for (const key in this.cachedInitialProject.files) {
+            await this.write(
+              key,
+              this.cachedInitialProject.files[key],
+              VersionAction.INCREMENT
+            );
           }
           this.host.notifyReady(this);
         } else {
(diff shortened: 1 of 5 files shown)


microbit-robert:
To do:

- [ ] Add tests for projects with multiple files in `host.test.ts`.
- [ ] Add tests for new methods in `fs.test.ts`.
- [ ] Update `statistics` method in `fs.ts` to handle multiple file projects.

src/fs/initial-project.ts
```diff
@@ -27,6 +29,21 @@ export interface InitialProject {
...
+
+export const isSimplePythonProject = (
+  project: any
+): project is SimplePythonProject => {
+  return (
+    project &&
+    project.hasOwnProperty("main") &&
+    typeof project.main === "string"
+  );
+};
```
microbit-matt-hillsdon, line 45: Is this a format we need to be backwards compatible with?
V2 editor uses a string, I think?

microbit-robert: You are right, so we now receive a string or (`MultiFilePythonProject`) object from the host. If it's a string, it's resolved to this `SimplePythonProject` before being passed into the file system `resolve({ isDefault: false, main: data.projects[0] });`

src/fs/host.ts
```diff
@@ -74,10 +74,12 @@ export class IframeHost implements Host {
               "'projects' array should contain at least one item."
             );
           }
-          if (typeof data.projects[0] !== "string") {
-            throw new Error("Expected string");
+          if (typeof data.projects[0] === "string") {
+            resolve({ isDefault: false, main: data.projects[0] });
+          }
+          if (typeof data.projects[0] === "object") {
+            resolve(data.projects[0]);
```
microbit-robert, line 81: Is this basic type checking good enough? Applies to `handleImportProject` below as well.

src/fs/initial-project.ts
microbit-robert: I've had a go at this here - [`7539a94`](https://github.com/microbit-foundation/python-editor-next/pull/626/commits/7539a94c8862618dcda7e690db1ddcaf3d5ebd3b). The `isDefault` field only seems to be used for testing, but I've kept it in for now.

microbit-matt-hillsdon: Ooh, I think isDefault was only used for the stats code we used the other day and can probably be removed now.

src/fs/fs.ts
```diff
@@ -320,6 +322,39 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
...
+  async getPythonProject(): Promise<PythonProject> {
+    const projectName = await this.storage.projectName();
+    const project: PythonProject = {
+      files: {},
+      projectName,
+    };
+    for (const file of await this.storage.ls()) {
+      const data = await this.storage.read(file);
+      const content = new TextDecoder().decode(data);
+      project.files[file] = content;
```
microbit-matt-hillsdon, line 334: I think we don't want to be doing this if we want binary file support.
Rather than taking the bytes and converting to a string we want to go bytes -> base64 (and in reverse when reading from a Python project).

src/fs/fs.ts
```diff
@@ -320,6 +322,39 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
...
+    this._dirty = false;
+    this.project = {
+      ...this.project,
+      id: generateId(),
+    };
+    if (project.projectName) {
+      await this.storage.setProjectName(project.projectName);
+    }
+    await this.overwriteStorageWithFs();
+    return this.notify();
```
microbit-matt-hillsdon, line 355: Maybe this bit can be in common?

src/fs/initial-project.ts
microbit-robert: Taken care of here - [`833c48a`](https://github.com/microbit-foundation/python-editor-next/pull/626/commits/833c48ad7f6adde6ef226f93025f9f59f3c3b27f)

src/fs/fs.ts
microbit-robert: File content encoded to base64 - [`2a6ae34`](https://github.com/microbit-foundation/python-editor-next/pull/626/commits/2a6ae34c0d7cd2b57103434db58ee04b5c850639)

src/fs/fs.ts
microbit-robert: [`d3b5ca7`](https://github.com/microbit-foundation/python-editor-next/pull/626/commits/d3b5ca73f3f81cfdde7cd50e1effdd04f019cd73)

src/fs/fs.ts
```diff
@@ -364,7 +394,7 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
       storageUsed: fs.getStorageUsed(),
       lines:
         this.cachedInitialProject &&
-        this.cachedInitialProject.main === currentMainFile
+        this.cachedInitialProject.files[MAIN_FILE] === btoa(currentMainFile)
```
microbit-matt-hillsdon, line 397: I'm worried about btoa with unicode (try it and see). Can we use readBytes and then the base64 code we use in storage and similar elsewhere?

microbit-robert: It would have to be something like this then?
`fromByteArray(new TextEncoder().encode(currentMainFile))`

`currentMainFile` is expected as a string in the ternary expression to determine the number of lines.

microbit-matt-hillsdon: Hmm. I suppose you could count the lines off the byte array, but not so concisely.

microbit-robert: Changes as discussed - [`6889191`](https://github.com/microbit-foundation/python-editor-next/pull/626/commits/688919105a41b5f82e2720ee850d92f342aeeafc).

src/fs/fs.ts
```diff
@@ -358,18 +388,33 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
...
+    const LF = 10;
+    const CR = 13;
+    arr.forEach((byte) => {
+      if ((byte === LF && prevByte !== CR) || byte === CR) {
+        lineCount++;
+      }
+      prevByte = byte;
+    });
+    return lineCount;
+  }
```
microbit-matt-hillsdon, line 416: I think I'd extract this to its own function and write a simple unit test for it, as it could easily be wrong and we wouldn't really notice. I think it's right though 😄

microbit-matt-hillsdon (approved):
Looks good. One comment about adding a test, you can add that separately if you prefer.

src/fs/fs.ts
microbit-robert: Extracted and test added here - [`23fcb8d`](https://github.com/microbit-foundation/python-editor-next/pull/626/commits/23fcb8d67ddb9653d8d439a87b42c76a59853112)

microbit-matt-hillsdon: Looks good, thanks.

PR #630, opened 2022-03-21
Make Connect the primary action.

microbit-matt-hillsdon:
Once connected, Flash is the primary action.
If WebUSB is not supported, Download is the primary action (as before
the change).

### DIFF
--- a/src/project/DownloadFlashButton.tsx
+++ b/src/project/DownloadFlashButton.tsx
@@ -35,8 +35,11 @@ interface DownloadFlashButtonProps {
  * flash (if WebUSB is supported) or otherwise just download a HEX.
  */
 const DownloadFlashButton = ({ size }: DownloadFlashButtonProps) => {
-  const connectionStatus = useConnectionStatus();
-  const connected = connectionStatus === ConnectionStatus.CONNECTED;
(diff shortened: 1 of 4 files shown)


src/project/DownloadFlashButton.tsx
```diff
@@ -35,21 +35,37 @@ interface DownloadFlashButtonProps {
...
+              size={size}
+            />
           ) : (
-            <DownloadButton width={buttonWidth} mode={"button"} size={size} />
+            <DownloadButton
+              variant={variant}
+              width={buttonWidth}
+              mode={"button"}
+              size={size}
+              borderRight="1px"
```
microbit-robert, line 62:
```suggestion
              borderRight={variant ? "0px" : "1px"}
```
This change keeps the gap consistent between the download/flash button and its additional actions consistent where WebUSB is not available. I.e., narrows the gap observed below to what it was previously.

image

microbit-robert (approved):
Minor suggestion added. Otherwise, this looks good.

src/project/DownloadFlashButton.tsx
microbit-matt-hillsdon: Good catch, thank you.

PR #634, opened 2022-03-25
Add error markers in the gutter with a friendlier rendering while the line is being edited

microbit-robert:
Requires us to inline CM's MIT licensed lint extension.

Once the dust has settled from this feature we'll do some work to see if we can propose a way of reusing more of CM's code here. There isn't an obvious seam between the gutter and the rest of the lint module, but perhaps one can be found. I think we may change this significantly, so we should wait a few weeks.


microbit-robert:
Source code for the lintGutter extension (useful reference if we want to develop our own more customisable version):
https://github.com/codemirror/lint/blob/084c4ad59be7562c739d2e8fff12c7852e106b15/src/lint.ts#L616-L632

Other examples of gutter markers:
https://codemirror.net/6/examples/gutter/

microbit-robert:
https://review-python-editor-next.microbit.org/gutter-errors/

src/editor/codemirror/lint/editingLine.ts
```diff
@@ -4,30 +4,37 @@ import { ViewPlugin, ViewUpdate } from "@codemirror/view";
...
       const mainIndex = update.state.selection?.asSingle()?.ranges[0]?.from;
-      console.log(update.state.selection);
       if (!mainIndex) {
         return undefined;
       }
-      console.log({ mainIndex });
       const doc = update.state.doc;
       const selectionLine = doc.lineAt(mainIndex).number;
+      // Don't update if currentlyEditingLine hasn't changed.
+      if (this.prevSelectionLine === selectionLine) {
```
microbit-matt-hillsdon, line 20: Let's use view.state.field(currentlyEditingLine)

microbit-robert: [`c04c1f1`](https://github.com/microbit-foundation/python-editor-next/pull/634/commits/c04c1f113320fd3b3a5f93667ff191a6a081b915)

microbit-robert:
Current problems:
image
Occurs when calling `markersForDiagnostics` inside the `if (effect.is(setEditingLineEffect))` statement.

Typing indicator is only present when there is also a diagnostic associated with the line being edited.

jaustin:
The latest setup where there's an 'unfilled' circle when editing is really good. 

Is the delay between leaving the line and the gutter indicator turning red the time it takes for the error to be generated, or is that a typing timeout? If we revisit this I think it'd be neat for leaving the line to immediately paint errors red (versus a timeout if you're still on the line) but it is not significant right now.

microbit-robert:
> The latest setup where there's an 'unfilled' circle when editing is really good.
> 
> Is the delay between leaving the line and the gutter indicator turning red the time it takes for the error to be generated, or is that a typing timeout? If we revisit this I think it'd be neat for leaving the line to immediately paint errors red (versus a timeout if you're still on the line) but it is not significant right now.

I agree that the gutter icon should change immediately on leaving the line. It's an issue with line edits vs selection changes triggering updates rather than a delay in the error being generated, but we'll look at that when we come back to this.

jaustin:
Thanks @microbit-robert  - I'd say from a feature completeness perspective this is good to merge from my PoV. 

I'm also beginning to wonder if the 'typing indicator' only being present when there's a diagnostic is really a feature, not a bug?

microbit-robert:
The markers are mapped through changes -> `markers = markers.map(tr.changes)`.
The diagnostics used to create the markers are not. Thus, we re-create the lint gutter markers using out of date diagnostics.

microbit-robert:
We only need to map the `from` field on diagnostics to recreate the gutter markers and we can get this from `iter.from` - see [`a7da9bf`](https://github.com/microbit-foundation/python-editor-next/pull/634/commits/a7da9bf8519b67aaad5cf2a214ae9d65692dcc18)

microbit-matt-hillsdon (approved):

PR #636, opened 2022-03-28
Rename explore and reference tabs

microbit-robert:
Rename Explore and Reference tabs to Reference (uses reference icon) and API (uses Python icon), respectively.


microbit-matt-hillsdon:
This looks good.

Please can you change the query to match slug.current with the values "explore" **or** "reference" ?
Then I'll update the CMS value and eventually drop the `or`.

microbit-robert:
> This looks good.
> 
> Please can you change the query to match slug.current with the values "explore" **or** "reference" ? Then I'll update the CMS value and eventually drop the `or`.

Sounds sensible, please see [`18b2bf7`](https://github.com/microbit-foundation/python-editor-next/pull/636/commits/18b2bf70bd7531e51ee2462a6ced3c5c3dca0c2a).

microbit-matt-hillsdon (approved):

PR #646, opened 2022-04-05
Beta prep

### LINKED ISSUE #669: Updates and impact on metrics
Good point raised by @microbit-carlos below. Discussion required on the best direction to take.

> Once we release V3 as beta, how often would we expect to deploy with updates?
The metrics currently create a new category based on the Editor version, so right now in alpha each category only contains a few days of data.
In a slow release schedule, like what we had with V2, that has worked fine, but maybe we want to change it in V3? As an option, we could only create a new category when we make changes to the metrics (or new UI widgets are added, which might add new metrics)?

microbit-matt-hillsdon:
- Changes the alpha release notice to a beta one.
- Bumps the theme pages to a version with the simple fix that closes #669.

### DIFF
--- a/src/workbench/InfoDialog.tsx
+++ b/src/workbench/InfoDialog.tsx
@@ -96,7 +96,7 @@ const InfoDialog = ({
                     },
                   }}
                 >
-                  Documentation
+                  Guidance
                 </Button>
               </Flex>
               <Text fontSize="sm" pt={8}>
--- a/src/workbench/InfoDialog.tsx
+++ b/src/workbench/InfoDialog.tsx
@@ -61,7 +61,7 @@ const InfoDialog = ({
                     Drag and drop them into your code.
                   </ListItem>
                   <ListItem m={1}>
-                    Autocomplete and paramerer help. See the choices available
+                    Autocomplete and parameter help. See the choices available
                     to you as you type.
                   </ListItem>
                   <ListItem m={1}>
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
Link directly to the "Guide" page and label it "Guidance". The Help & support link remains the same.
MicroPython Documentation -> MicroPython documentation for consistency.

src/workbench/InfoDialog.tsx
```diff
@@ -36,58 +47,64 @@ const InfoDialog = ({
...
                   <ListItem m={1}>
-                    Autocomplete and signature help in the code editor.
+                    Working code examples in the <strong>Reference</strong> tab.
+                    Drag and drop them into your code.
                   </ListItem>
                   <ListItem m={1}>
-                    Drag and drop code examples from the{" "}
-                    <strong>Reference</strong> and <strong>API</strong>{" "}
-                    documentation tabs.
+                    Autocomplete and paramerer help. See the choices available
```
microbit-robert, line 64:
```suggestion
                    Autocomplete and parameter help. See the choices available
```

microbit-robert (changes requested):
See suggestion for typo fix. Otherwise this looks good.

microbit-matt-hillsdon:
> See suggestion for typo fix. Otherwise this looks good.

Good catch, thanks @microbit-robert.

PR #647, opened 2022-04-06
Missing aria-label.

microbit-matt-hillsdon:
Make accessibility tree in Chrome easier to review by removing dialog
that's not present on screen.


microbit-robert (approved):
LGTM.

Issue #638, opened 2022-03-28
Sidebar in a bad state


microbit-matt-hillsdon:
![4FB68846-EE91-4C3C-8A86-B1D9BC7D40CE](https://user-images.githubusercontent.com/44397098/160454645-5e499675-d437-47e9-a01e-8d874a8f7f5e.jpg)

From @jaustin:

At this point the divider was also undraggable.
FWIW I had dev tools on the right of the editor when I got to this state.

jaustin:
A confirmed way to reproduce this is to set the window to 704 pixels (or about that) and then drag the sidebar out to the right (IE try to enlarge it). It immediately snaps to 'broken size' and cannot be recovered.

PR #648, opened 2022-04-06
Enforce the minimum size of the sized component.

microbit-matt-hillsdon:
Fixes https://github.com/microbit-foundation/python-editor-next/issues/638

@microbit-robert, please try this carefully at different window sizes.


microbit-robert:
@microbit-matt-hillsdon This seems to fix the referenced issue, however, gives rise to a window wide horizontal scrollbar previously avoided which can lead to some interesting visuals when scrolled all the way to the right.

image

microbit-robert (approved):
LGTM.

PR #649, opened 2022-04-06
Shrink buttons on project action bar when narrow

microbit-matt-hillsdon:
Needs to use a different width that isn't affected by padding to avoid
amusing effects in the intersection.

### DIFF
--- a/src/common/use-resize-observer.ts
+++ b/src/common/use-resize-observer.ts
@@ -10,6 +10,7 @@ export const useResizeObserverContentRect = (
     const resizeObserver = new ResizeObserver((entries) => {
       entries.forEach((entry) => {
         if (entry.contentRect) {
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
If folks like this then I'd like to pull it up, to the whole of the main area, and also adapt the right spacing of the divider at the top so it continues to match the right edge of the load button. Then in future e.g. we might hide the gutter.

microbit-matt-hillsdon:
If folks like this then I'd like to pull it up, to the whole of the main area, and also adapt the right spacing of the divider at the top so it continues to match the right edge of the load button. Then in future e.g. we might hide the gutter.

microbit-matt-hillsdon:
We shipped some padding changes but decided against changing button size. I think we're more likely to consolidate Save/Open buttons in any future work here. Closing PR.

PR #661, opened 2022-04-07
Remove CM workaround that's no longer needed.

microbit-matt-hillsdon:
The sandbox on my CM discuss post is now fixed and we've long since
upgraded:
https://discuss.codemirror.net/t/diagnostics-for-the-range-from-end-of-a-line-to-the-start-of-the-next/3495


microbit-robert (approved):
LGTM. No change in behaviour observed.

PR #664, opened 2022-04-07
Files -> Project tab

microbit-matt-hillsdon:
Use the project name as a heading to make it clearer that the files
belong to the project you're editing, rather than each file being a
project. This addresses user testing feedback.
   
Prefer "Add" terminology to "Create" to make it clearer the file is
added to the project, not creating anything new.


microbit-robert (approved):
LGTM.

Issue #663, opened 2022-04-07
Minimal "Ideas" section for beta


microbit-matt-hillsdon:
## Depends on

Content from @microbit-giles, itself blocked on the CMS structure for this.

## Description

Content model:

- A title
- A photo-like image
- V1/V2 support tag info similar to current toolkit content
- Rich text with code blocks, generally containing a single code example

UI will use a two level design similar to Reference but the top and second levels should make good use of the image.

Top level will show only image and title.

Second level will show all content with a breadcrumb nav back to the top level.

## Todo

- [x] CMS structure
- [x] Initial content (can be dev-created)
- [x] Good enough for alpha content from @microbit-giles 
- [x] Update readme

jaustin:
This won't be translated initially as the translation has started, but as there are a small number of strings here, perhaps we can include it in a second round?

microbit-matt-hillsdon:
Raised #672 for Ideas search.

I'll follow up separately on plans for translation as this has missed the window for the French for beta.

PR #666, opened 2022-04-08
Ideas tab content

microbit-robert:
Initial implementation of ideas cards and content.

See #663.


microbit-robert:
Todo:
- [ ] Fix issue with api / query returning 0 results. Query runs successfully in the CMS. Data is currently hardcoded for demo purposes.
- [ ] Adjust design after feedback. Currently unsure about general design, and specifically `flex-wrap` / `max-width` properties.

PR #667, opened 2022-04-11
Add clear serial shortcut and menu item

### LINKED ISSUE #527: Clear serial action
From @microbit-katie:
> Is there a way to clear serial without disconnecting and reconnecting the micro:bit?

image

There isn't currently a way to do this but it should be straightforward to add, if we're happy to add to this menu:

image

microbit-robert:
Closes #527

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -471,6 +471,10 @@
     "defaultMessage": "Send Ctrl+D to reset",
     "description": "Button to reset the micro:bit from the serial area"
   },
+  "serial-ctrl-shift-l-action": {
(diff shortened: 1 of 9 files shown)


microbit-robert:
Let's review the shortcut UI string as well as the shortcut command. It is currently `Ctrl+Shift+L` so it doesn't interfere with with navigating to the browser URL/search bar, but it could overwrite this shortcut if we wanted it to.

microbit-matt-hillsdon:
This isn't high enough value for us to put in at this point in the release. Closing.

PR #668, opened 2022-04-11
Add support for language code on URL

### LINKED ISSUE #536: Support language code on URL as per V2 editor
Particularly important for links from microbit.org, the V2 editor supports specifying a language on the URL.

microbit-robert:
Intentionally supports the same l=${code} syntax as the V2 editor to help keep links elsewhere working.

Closes #536

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -43,6 +43,10 @@ interface Options {
    * URL fragment including the #.
    */
   fragment?: string;
+  /**
+   * Language parameter passed via URL.
+   */
+  language?: string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


microbit-robert:
Discuss resolving the clash between using the `languageId` from the URL vs recovering a `languageId` from `localStorage`.

microbit-matt-hillsdon:
I think this has the right behaviour: a past choice via the language dialog is respected over the value on the URL.

I think that behaviour is subtle enough to deserve a e2e test. Other than flags we don't have anything that faffs with the URL, I'll have a look at the best way to add this to app.ts.

PR #674, opened 2022-04-13
Fix logged react errors in dev mode.

microbit-matt-hillsdon:
- It's easy to end up with undefined content from the CMS so cope with
   that at a low level.
 - Fix a DOM nesting issue.


microbit-robert (approved):
LGTM.

PR #675, opened 2022-04-13
Add basic e2e test for the ideas tab

microbit-robert:
Adds e2e tests to check navigation to ideas tab, and selection of an individual idea.

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -494,13 +494,15 @@ export class App {
 
   async findDocumentationTopLevelHeading(
     title: string,
-    description: string
+    description?: string
   ): Promise<void> {
     const document = await this.document();
     await document.findByText(title, {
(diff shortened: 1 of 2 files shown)


src/e2e/app.ts
```diff
@@ -511,6 +513,17 @@ export class App {
...
 
+  async selectDocumentationIdea(name: string): Promise<void> {
+    const document = await this.document();
+    const heading = await document.findByText(name, {
+      selector: "h3",
+    });
+    const handle = heading.asElement();
+    await handle!.evaluate((element) => {
+      element.parentElement?.click();
+    });
```
microbit-matt-hillsdon, line 524: It feels like there should be a better way but I failed to make one work in a reasonable amount of time.

microbit-matt-hillsdon (approved):

PR #676, opened 2022-04-13
Add V2 tag to individual ideas header

### LINKED ISSUE #673: [Ideas] Display V2 tag in ideas header not just the cards
Screenshot2022_04_13_142437

We don’t show it on the title for reference but we show it entry by entry:

microbit-robert:
Closes #673.


microbit-matt-hillsdon (approved):
LGTM.

PR #678, opened 2022-04-13
Update typeshed to collapse display.show overloads.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
e2e test failures are due to this change, ignoring them for now until I've discussed this more widely.

PR #681, opened 2022-04-14
Collapse display.show overloads to correspond to reality

microbit-matt-hillsdon:
Also fixes display.scroll parameter name.


microbit-robert (approved):
LGTM.

PR #682, opened 2022-04-14
Remove unused content.

microbit-matt-hillsdon:
This has since moved to InfoDialog and been edited for beta.


microbit-robert (approved):
LGTM.

PR #685, opened 2022-04-18
Bump notice version for alpha -> beta.

microbit-matt-hillsdon:
So folks who previously saw the alpha notice see the beta one. At least initially the domain will be the same.


microbit-robert (approved):
LGTM.

PR #699, opened 2022-04-22
Fix boot event.

microbit-matt-hillsdon:
Should be type not action, not spotted as in this one scenario (not
using the hook) deployment was typed `any`.


microbit-matt-hillsdon:
I tested the event locally as it won't be triggered on the review branch.

microbit-robert (approved):
LGTM.

PR #702, opened 2022-04-22
Add connect assistance dialogs

microbit-robert:
Add dialogs to assist users in connected their micro:bit via WebUSB.

Dialogs added:
- Connect help
- Micro:bit not found
- Firmware update


microbit-robert:
- Finalised UI strings need adding properly to lang/en.json etc.

microbit-matt-hillsdon:
What should the Safari/Firefox experience be? You currently get the dialog explaining how it works in Chrome followed by the error toast.

Do we need a different GIF for Edge?

microbit-robert:
> What should the Safari/Firefox experience be? You currently get the dialog explaining how it works in Chrome followed by the error toast.
> 
> Do we need a different GIF for Edge?

The previous behaviour in Safari/Firefox is now preserved [`03a14ac`](https://github.com/microbit-foundation/python-editor-next/pull/702/commits/03a14ac06a9c7d8e99e986463005390fa44e781d). We can ask for feedback on whether this needs to be changed when we demo this branch?

What is different for Edge? My connect/cancel buttons are switched compared to the current GIF for both Edge and Chrome now that I really look at it.

src/common/use-dialogs.tsx
```diff
@@ -105,6 +123,19 @@ export class Dialogs {
       this.progressDialogSetState(options);
     }
   }
+
+  generic(options: GenericDialogParameters): Promise<boolean> {
```
microbit-matt-hillsdon, line 127: Can/should we use this to implement some of the others?

src/device/device.ts
```diff
@@ -513,6 +518,12 @@ export class MicrobitWebUSBConnection
     }
   };
 
+  clearDevice(): void {
+    this.connection = undefined;
+    this.device = undefined;
```
microbit-matt-hillsdon, line 523: What should this do if you're connected? Make it async and disconnect first? 

We plan to extract this component at some point so its API needs to be reasonably defensive.

microbit-matt-hillsdon:
Let's also test connecting a device in maintenance mode.

microbit-robert:
[`f59f065`](https://github.com/microbit-foundation/python-editor-next/pull/702/commits/f59f0655e9e9b9189f435f498b85181e2b441234) covers the text changes in #718. The change updates existing UI strings.

src/device/device.ts
microbit-robert: [`ebdc644`](https://github.com/microbit-foundation/python-editor-next/pull/702/commits/ebdc644bd0769de357211510a6da495abc6f207b)

lang/en.json
```diff
@@ -240,7 +276,7 @@
     "description": "Progress dialog text for partial flashes (quick)"
   },
   "flashing-full-flash-detail": {
-    "defaultMessage": "Initial flash might take longer, subsequent flashes will be quicker.",
+    "defaultMessage": "Your code is being sent ('flashed') to your micro:bit. This can take a while the first time, sorry, but it will be quicker after that.",
```
microbit-matt-hillsdon, line 279: I might look at temporarily reinstating the current French translation for this one.

src/project/project-actions.tsx
```diff
@@ -476,18 +554,50 @@ export class ProjectActions {
...
+          this.device.clearDevice();
+          // Temporary hide for French language users.
+          if (this.settings.values.languageId !== "en") {
+            return this.actionFeedback.expectedError(
+              this.webusbErrorMessage(e.code)
+            );
+          }
+          return await this.dialogs.show<void>((callback) => (
+            <FirmwareDialog callback={callback} />
+          ));
```
microbit-matt-hillsdon, line 600: Let's pull this out into a method similar to handleNotFound for consistency.

microbit-matt-hillsdon (changes requested):
This is looking good. Just e2e + one suggested tweak.

src/project/project-actions.tsx
microbit-robert: [`f012147`](https://github.com/microbit-foundation/python-editor-next/pull/702/commits/f0121472c8ef006e93cb31b5f9f675aab6d73f65)

lang/en.json
microbit-matt-hillsdon: We can just defer pulling French back from Crowdin for now.

microbit-matt-hillsdon:
image

This link looks like it might contain the wrong text.

Please can we also `vertical-align: middle` the external link icons?

microbit-matt-hillsdon (approved):
This is looking good now.

As discussed, there are a couple of things that I'd like to improve but I think we should tackle separately:

1. We're short on space for larger dialogs, so we've centred them rather than the usual fixed offset. This looks a bit odd on taller screens. No quick CSS fix so let's look at it separately.

2. The WebUSB not supported dialog uses the same text as the toast it replaces, but it seems significantly less friendly in a dialog. I'd like to improve this, but will spin out as we need to consider translation.

microbit-matt-hillsdon:
Raised https://github.com/microbit-foundation/python-editor-next/issues/729 and https://github.com/microbit-foundation/python-editor-next/issues/730

PR #703, opened 2022-04-22
Remove Box component affecting spacing

microbit-robert:
See Variable -> Changing number variables for example.


microbit-matt-hillsdon (approved):
Good catch, thanks.

PR #704, opened 2022-04-22
Edit position and style of 'show more'

### LINKED ISSUE #688: [Toolkits] Show more button and content layout.
When activating a `show more` button, the focus on the button is retained and text is displayed above the button rather than below it. This could be problematic for users using screen readers.

microbit-robert:
Show more/less now stays above the expanded/collapsed content, and has been re-styled to match other hyperlinks in the present in the toolkits.

Closes #688


microbit-matt-hillsdon:
This feels much less weird. Let's get internal feedback on the UI change.

src/documentation/common/ShowMoreLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+import { Link, LinkProps } from "@chakra-ui/react";
+import { FormattedMessage } from "react-intl";
+
+interface ShowMoreLinkProps extends LinkProps {
+  isOpen: boolean;
+}
+
+const ShowMoreLink = ({ isOpen, ...props }: ShowMoreLinkProps) => (
+  <Link {...props} color="brand.600">
```
microbit-matt-hillsdon, line 14: I think it should still actually be a button (maybe via `as="button"`).

microbit-matt-hillsdon (changes requested):
One comment inline, otherwise LGTM, including the contrast of the text.

There's a conflict to fix.

src/documentation/common/ShowMoreLink.tsx
microbit-robert: I've also changed the name back to `ShowMoreButton` to match - [`ef4d9e9`](https://github.com/microbit-foundation/python-editor-next/pull/704/commits/ef4d9e90f093b05b01029fefe1a76bc2d874c716)

microbit-matt-hillsdon (approved):
Agreed to change back to the darker purple.
Otherwise LGTM, please merge after colour change.

PR #706, opened 2022-04-22
Add missing copyright headers.

microbit-matt-hillsdon:


src/editor/codemirror/lint/lint.ts
```diff
@@ -1,4 +1,4 @@
-/*
```
microbit-matt-hillsdon, line 1: So the header is not appended here.

PR #710, opened 2022-04-25
Fix arrow alignment issue on show-more button in the API toolkit

microbit-robert:
Resolves the `show more` button alignment issue when an API entry is hovered (the arrow associated with the button moves on hover).


microbit-matt-hillsdon:
I'd spotted this but convinced myself I could only repro it on my WIP branch not on main. Glad I was wrong as I was very confused!

PR #711, opened 2022-04-27
Make code examples focusable

### LINKED ISSUE #700: [Toolkits] Code examples should be focusable
Code examples in the toolkits should be focusable to allow a user to copy the code (or perhaps use `enter` to insert the code).

microbit-robert:
Make code examples focusable. After the element is focused: 
- Use `ctrl + c` to copy text inside the code example. This copies the text shown (does not include imports, includes signatures in the API toolkit).
- Use `Enter` to insert the code into the editor. This works in the same way as the insert code button does for the Reference toolkit.

Closes #700.

### DIFF
--- a/src/documentation/common/CodeEmbed.tsx
+++ b/src/documentation/common/CodeEmbed.tsx
@@ -71,7 +71,7 @@ const CodeEmbed = ({ code: codeWithImports, parentSlug }: CodeEmbedProps) => {
 
   const actions = useActiveEditorActions();
   const handleInsertCode = useCallback(
-    () => actions?.insertCode(codeWithImports, parentSlug),
+    () => actions?.insertCode(codeWithImports, "example", parentSlug),
     [actions, codeWithImports, parentSlug]
   );
 
--- a/src/documentation/common/CodeEmbed.tsx
+++ b/src/documentation/common/CodeEmbed.tsx
@@ -95,7 +95,20 @@ const CodeEmbed = ({ code: codeWithImports, parentSlug }: CodeEmbedProps) => {
   const textHeight = lineCount * 1.375 + "em";
   const codeHeight = `calc(${textHeight} + var(--chakra-space-2) + var(--chakra-space-2))`;
   const codePopUpHeight = `calc(${codeHeight} + 2px)`; // Account for border.
-
+  const isMac = /Mac/.test(navigator.platform);
+  const handleKeyDown = useCallback(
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
When I drag and drop from API, the cursor is placed between the function brackets. When I insert the code via the new keyboard interaction the cursor is placed after the brackets. It feels like these should behave the same and use essentially the same code path.

microbit-matt-hillsdon:
When I drag and drop from API, the cursor is placed between the function brackets. When I insert the code via the new keyboard interaction the cursor is placed after the brackets. It feels like these should behave the same way, which is slightly different from Reference.

microbit-robert:
> When I drag and drop from API, the cursor is placed between the function brackets. When I insert the code via the new keyboard interaction the cursor is placed after the brackets. It feels like these should behave the same way, which is slightly different from Reference.

This should align the insert and dnd behaviour - [`231b6d0`](https://github.com/microbit-foundation/python-editor-next/pull/711/commits/231b6d07bb7e680019ca44c50e060cfc61f6fea6).

microbit-robert:
To successfully add an `aria-label` that is read by screen readers, we would need to explicitly define a role, none of which really seem appropriate for this.

microbit-matt-hillsdon:
> To successfully add an `aria-label` that is read by screen readers, we would need to explicitly define a role, none of which really seem appropriate for this.

Maybe we do visually hidden text inside what's selected (but outside of CM) ?

microbit-matt-hillsdon (approved):
Code change looks good. I'd like us to spend some time thinking about common code for keyboard shortcuts next time we do something similar.

Worth trying visually hidden text as per my other comment but can be follow-up PR if you prefer.

src/documentation/common/CodeEmbed.tsx
```diff
@@ -109,6 +122,15 @@ const CodeEmbed = ({ code: codeWithImports, parentSlug }: CodeEmbedProps) => {
...
           highlightDragHandle={state === "raised"}
+          tabIndex={0}
+          _focus={{
+            boxShadow: "var(--chakra-shadows-outline);",
+          }}
+          _focusVisible={{
+            outline: "none",
+          }}
+          onKeyDown={handleKeyDown}
+          zIndex={1}
```
microbit-matt-hillsdon, line 133: I blame this for https://github.com/microbit-foundation/python-editor-next/issues/727

PR #722, opened 2022-05-03
Simplify dialogs API. Remove actions from dialogs.

microbit-matt-hillsdon:
Targets in-progress connect-dialogs branch.

### DIFF
--- a/src/project/project-actions.tsx
+++ b/src/project/project-actions.tsx
@@ -114,7 +114,7 @@ export class ProjectActions {
   /**
    * Show connection help with options to connect or cancel.
    *
-   * @param force true to show the help even if the user previously requestede not to (used in error handling scenarios).
+   * @param force true to show the help even if the user previously requested not to (used in error handling scenarios).
    * @return true to continue to connect, false to cancel.
    */
   private async showConnectHelp(force: boolean): Promise<boolean> {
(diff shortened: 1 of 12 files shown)


src/project/project-actions.tsx
```diff
@@ -93,42 +94,69 @@ export class ProjectActions {
...
+        await this.connectInternal();
+      }
     }
   };
 
   /**
-   * Connect to the device if possible, otherwise show feedback.
+   * Show connection help with options to connect or cancel.
+   *
+   * @param force true to show the help even if the user previously requestede not to (used in error handling scenarios).
```
microbit-robert, line 117:
```suggestion
   * @param force true to show the help even if the user previously requested not to (used in error handling scenarios).
```

src/project/project-actions.tsx
```diff
@@ -156,7 +184,8 @@ export class ProjectActions {
    * of modules may be opened together. The existing project is
    * updated.
    *
-   * @param file the file from drag and drop or an input element.
+   * @param files the files from drag and drop or an input element.
+   * @param type the type of user event that triggered the load.
```
microbit-robert, line 188:
```suggestion
   * @param the type of user event that triggered the load.
```

microbit-robert (approved):
LGTM. This is much simpler.

PR #724, opened 2022-05-04
Update for machine.mem{8,16,32} addition.

### LINKED ISSUE #723: machine.mem{8,16,32} missing from stubs
Raised via the feedback form:

image
>Missing machine.mem as used in some example code from Kitronik

This is just straightforwardly missing from the stubs. It's documented but as prose so it was missed: https://microbit-micropython.readthedocs.io/en/v2-docs/machine.html#reading-memory

microbit-matt-hillsdon:
This also includes another minor stubs fix: removing the second argument to sys.print_exception which isn't supported for micro:bit MicroPython.

Closes #723


microbit-matt-hillsdon:
This is looking good to me. I'm going to go ahead and merge.

PR #726, opened 2022-05-04
Remove unused structure highlighting modes.

microbit-matt-hillsdon:
These were kept for experimentation but are now a barrier to
understanding and changing this code.

### DIFF
--- a/src/editor/codemirror/structure-highlighting/view.ts
+++ b/src/editor/codemirror/structure-highlighting/view.ts
@@ -50,8 +50,7 @@ export const codeStructureView = (option: "full" | "simple") =>
       }
 
       update(update: ViewUpdate) {
-        // We can probably limit this but we need to know when the language state has changed
-        // as parsing has occurred. We also need to be updated when hint state changes.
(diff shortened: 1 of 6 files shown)


src/editor/codemirror/structure-highlighting/view.ts
```diff
@@ -46,26 +44,14 @@ export const codeStructureView = (settings: CodeStructureSettings) =>
...
-        );
+        this.overlayLayer.classList.add("cm-cs--mode-" + option);
         this.overlayLayer.setAttribute("aria-hidden", "true");
         view.requestMeasure(this.measureReq);
       }
 
       update(update: ViewUpdate) {
-        // We can probably limit this but we need to know when the language state has changed as parsing has occurred.
+        // We can probably limit this but we need to know when the language state has changed
+        // as parsing has occurred. We also need to be updated when hint state changes.
```
microbit-matt-hillsdon, line 54:
```suggestion
        // We can probably limit this but we need to know when the language state has changed as parsing has occurred.
```

microbit-robert (approved):
LGTM.

PR #733, opened 2022-05-06
Fix for RangeError

microbit-robert:
Fix for `RangeError` observed in dev tools when running the project locally. Occurs when there is no `lintState` field in the CodeMirror editor state.


microbit-matt-hillsdon (approved):
Fix looks good to me.

I think the lint module might only activate itself when you first set diagnostics, which I'd forgotten so it makes sense.

Issue #677, opened 2022-04-13
[Serial] Terminal hits error and stays in broken state until refresh


microbit-matt-hillsdon:
Giles got into a state where he could not interact with the REPL or see serial output.

This persisted until we refreshed the tab regardless of disconnect/reconnect/flash.

I think it's represented by [this Sentry issue](https://sentry.io/organizations/microbit/issues/3190440968/?project=5677402&query=is%3Aunresolved). There's a similar one with working source map [here](https://sentry.io/organizations/microbit/issues/3176991175/?project=5677402&query=is%3Aunresolved+_forEachCell&statsPeriod=14d).

microbit-matt-hillsdon:
This is the same as https://github.com/microbit-foundation/python-editor-next/issues/603. I also saw it on a call with Jonny earlier today.

Logs from #603:

```
instrument.ts:125 Device status CONNECTED
helpers.ts:120 Uncaught TypeError: Cannot read properties of undefined (reading 'loadCell')
    at t._forEachCell (xterm.js:1:122707)
    at t._drawBackground (xterm.js:1:123185)
    at t.onGridChanged (xterm.js:1:124882)
    at t.renderRows (xterm.js:1:116358)
    at t._renderRows (xterm.js:1:162731)
    at e._renderCallback (xterm.js:1:160756)
    at e._innerRefresh (xterm.js:1:36928)
    at xterm.js:1:36644
    at i (helpers.ts:98:17)
instrument.ts:125 Stopping serial before flash
```

microbit-matt-hillsdon:
You get the same trace if you open() a terminal twice, with or without an intervening dispose.

I think XTerm.js is being remounted (reason TBC). It does the open call but not clean-up as there's a weird split of responsibilities between the hook and the component. It's like that to support the serial actions having access to the terminal.

A related bug can be triggered by changing language after connect. That results in the terminal being disposed but not recreated (triggered by the value of intl changing).

I think we should change the design so that the context tracks the current terminal (probably via a ref) which is used for the actions and all the lifecycle management moves into the XTerm component.

PR #734, opened 2022-05-09
Fix terminal lifecycle issues.

microbit-matt-hillsdon:
This could cause issues if XTerm was remounted or when the user changed
language.

Closes https://github.com/microbit-foundation/python-editor-next/issues/677


src/messages/FixedTranslationProvider.tsx
```diff
@@ -3,9 +3,8 @@
  *
  * SPDX-License-Identifier: MIT
  */
-import { useSettings } from "../settings/settings";
-import { IntlProvider, MessageFormatElement } from "react-intl";
-import { ReactNode, useEffect, useState } from "react";
+import { ReactNode } from "react";
+import { IntlProvider } from "react-intl";
```
microbit-matt-hillsdon, line 7: Tangentially related lint issue - I used this class in the test harness to trigger the bug. Evidently its normal test use isn't linted. I've made a note to look into that.

microbit-matt-hillsdon:
Confirmed that this fixes the language switching bug.

I've also updated my test harness that remounted XTerm repeatedly and that now runs without error.

I don't understand why the remounts happen in the original scenarios in #677, but I think I'm happy enough to call it fixed on the basis that we've fixed a logical error that would explain the issue.

PR #737, opened 2022-05-09
Storage security error

### LINKED ISSUE #736: Handle SecurityError thrown from window.sessionStorage access
Happens in Safari 15.3 according to analytics and Chrome 101.

Neither straightforwardly reproduce for me so perhaps it depends on settings.

Seems fair enough to swallow the error as we have a fallback.

microbit-matt-hillsdon:
Closes #736


microbit-robert (approved):
LGTM.

Issue #740, opened 2022-05-11
[Stubs] Fix os.uname return value


microbit-matt-hillsdon:
We've typed os.uname's return value as a string but it is a richer value with extra detail.

Error in program that runs correctly:

<img width="496" alt="Screenshot 2022-05-11 at 05 47 01 (1)" src="https://user-images.githubusercontent.com/44397098/167803786-37c099a3-408b-40eb-a3b6-06057a195f0f.png">

raised via https://support.microbit.org/a/tickets/53361

REPL:
```
>>> import os
>>> os.uname()
(sysname='microbit', nodename='microbit', release='2.0.0', version='micro:bit v2.0.0+16d4380 on 2021-10-25; MicroPython 16d4380 on 2021-10-25', machine='micro:bit with nRF52833')
>>> type(os.uname())
<class 'tuple'>
>>> os.uname().version
'micro:bit v2.0.0+16d4380 on 2021-10-25; MicroPython 16d4380 on 2021-10-25'
>>> len(os.uname())
5
```

Potentially we can use the typeshed version: https://github.com/python/typeshed/blob/master/stdlib/os/__init__.pyi#L450

PR #743, opened 2022-05-11
Update stubs for os.uname.

microbit-matt-hillsdon:
Tweak dnd rules so you can drop uname_result fields.

Closes https://github.com/microbit-foundation/python-editor-next/issues/740


microbit-matt-hillsdon:
Stubs PR is merged. This now matches main and is good to go.

Issue #744, opened 2022-05-12
French translation: Chargement > Charger


microbit-mark:
Load button on the right should be an action verb in French as with other buttons. Suggest 'Charger'

<img width="795" alt="Screenshot 2022-05-12 at 11 45 56" src="https://user-images.githubusercontent.com/31242877/168053819-f90a431d-cabb-405c-88c0-91582f04ec46.png">

microbit-matt-hillsdon:
We should review the Crowdin context for this to see if it could be clearer for the benefit of future translation in other languages.

microbit-matt-hillsdon:
Checked with Steph and she agrees with the change.

CC @microbit-Emma for info. I'll change this directly in Crowdin.

PR #745, opened 2022-05-12
Fix Load button text in French.

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-next/issues/744


microbit-matt-hillsdon:
Fixed version, c.f. screenshot on #744.

image

PR #747, opened 2022-05-19
Improve common student error messages

microbit-robert:
Pyright build uses changes made in https://github.com/microbit-foundation/pyright/pull/11.
See that PR for the details of the changes.


microbit-matt-hillsdon:
@microbit-robert, I reviewed the error scenarios document and there are a few more to add to the common list. Missing:

- Inconsistent use of tabs and spaces in indentation (no change, just needed for translation)
- X is not defined  (no change, just needed for translation)
- Cannot access member "foo" for type "Bar"
    - We need to keep the type here I think, but we should probably drop the addendum.

I think we should revisit the "Unpacking more elements than returned" case too.

microbit-matt-hillsdon:
The issues in https://github.com/microbit-foundation/python-editor-next/pull/747#issuecomment-1133143864 have been addressed along with a few more tweaks.

microbit-matt-hillsdon (approved):
We'll do another round of review about the messages.

PR #751, opened 2022-05-24
Reinitialize Pyright on locale change

microbit-matt-hillsdon:
This gets us translated stubs and error messages, or it would, if we had any translations.

This puts the mechanism in place.


src/language-server/client.ts
```diff
@@ -110,6 +110,7 @@ export class LanguageServerClient extends EventEmitter {
       });
 
       const initializeParams: InitializeParams = {
+        locale: "fr",
```
microbit-matt-hillsdon, line 113: Obviously a hack!

bin/update-pyright.sh
```diff
@@ -8,8 +8,8 @@ set -euxo pipefail
 (cd ../pyright/packages/browser-pyright && npm run build)
 rm public/workers/pyright-*.js*
```
microbit-matt-hillsdon, line 9: This needs to also delete chunk files.

microbit-matt-hillsdon: It might be possible to set a webpack config option that will prefix the chunk file names so they're easier to delete here.

microbit-robert:
Todo:

- [ ] Update `update-typeshed.sh` script to account for multiple languages.

microbit-matt-hillsdon:
When we integrate the first French translations of the Pyright content we should definitely add an e2e test as it's the kind of thing we could break at multiple layers.

microbit-robert:
@microbit-matt-hillsdon This all looks good now. I updated the typeshed again after merging the stubs from main and rebuilding (we had some issues from urandom -> random etc.). I've created a PR for falling back to simplified English strings in Pyright https://github.com/microbit-foundation/pyright/pull/17. I think once this is merged and rebuilt, we can delete the French typeshed file and case and we're good to go.

microbit-robert (approved):
LGMT.

src/language-server/client.ts
```diff
@@ -170,6 +169,21 @@ export class LanguageServerClient extends EventEmitter {
     return this.initializePromise;
   }
 
+  async getInitializationOptions(): Promise<any> {
```
microbit-matt-hillsdon, line 172:
```suggestion
  private async getInitializationOptions(): Promise<any> {
```

Issue #754, opened 2022-05-27
[Search] Clearing search box does not always clear the results


microbit-robert:
To recreate, type "display" in the search box and wait for the results to populate. Hold backspace to clear the search input. Search results will be shown using the starting letter, in this case "d", of the previous search term. This seems to happen invariably, but the results are misleading. If you try it with "loop", it appears to work, but this is only because there are no results for the search term "l".

![image](https://user-images.githubusercontent.com/95928279/170668493-83c0b194-be85-4ab6-bedc-3e0139195c62.png)

This does not happen if the clear "x" button is used on the right side of the search box.

microbit-matt-hillsdon:
Good catch! Seems to be down to clearing the results outside of the debounced search function that can then be overwritten by the debounced (therefore delayed) change. Fixed on linked PR.

PR #757, opened 2022-05-27
Fix timing bug.

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-next/issues/754

The debounced change for removing the last character happens ~300ms
after we clear it outside of the debounce function. We can leave the
debounced function to handle this case.


microbit-robert (approved):
LGTM.

PR #763, opened 2022-06-01
[DO NOT MERGE] Demo of button re-ordering and connect flow.

microbit-robert:
Demo of button re-ordering and connect flow following user testing.


microbit-robert:
On switching from the cable connect dialog to the connect help dialog, the tooltip on the trigger button overlays the second modal. We can either change the focus on modal close, or merge these dialogs into one where the next button simply loads new content into the existing dialog.

microbit-robert:
If the project name is the default, the edit project name dialog is triggered. If the user cancels the name edit, should the file still be downloaded with the default name, or should it stop the download?

microbit-robert:
Todo:
- [ ] Add overlay on USB device selection.
- [ ] Merge connect cable and help dialog, add back button (wizard).
- [ ] If edit project name dialog is shown on download, the cancel button should also cancel the download as well as the dialog.
- [ ] Use "open" in open tooltip rather than "load".
- [ ] Review the use of "flash" in UI strings.
- [ ] Clicking try again on the 'not found' dialog should show connect help regardless of user settings.

microbit-robert:
Final points to review:
- [ ] The manually implemented time to wait between connect and flash, minimum seems to be 10ms.
- [ ] Send to micro:bit tooltip can cover dialogs on trying again from the 'not found'/'firmware' dialogs.
- [ ] On 'try again' we don't know if the user's intention was to connect or flash the micro:bit.

src/project/DownloadMenuButton.tsx
```diff
@@ -34,44 +31,26 @@ interface DownloadFlashButtonProps {
...
-              variant={variant}
-              width={buttonWidth}
-              mode={"button"}
-              size={size}
-              borderRight={variant ? "0px" : "1px"}
-            />
-          )}
+          <DownloadButton
+            width={buttonWidth}
+            mode={"button"}
```
microbit-matt-hillsdon, line 44: Odd JSX syntax

src/project/ProjectNameEditable.tsx
```diff
@@ -74,7 +49,7 @@ const ProjectNameEditable = ({
         fontSize="xl"
         color="brand.500"
         variant="ghost"
-        onClick={handleEdit}
+        onClick={() => actions.editProjectName()}
```
microbit-matt-hillsdon, line 52: You can pass actions.editProjectName directly (and avoid recreating a function each render). That's why they're arrow functions in ProjectActions.

src/project/ProjectNameEditable.tsx
```diff
@@ -83,7 +58,7 @@ const ProjectNameEditable = ({
     <Text
       key="text"
       cursor={clickToEdit ? "pointer" : undefined}
-      onClick={clickToEdit ? handleEdit : undefined}
+      onClick={() => (clickToEdit ? actions.editProjectName() : undefined)}
```
microbit-matt-hillsdon, line 61: Can this be `{clickToEdit ? actions.editProjectName : undefined}`

?

src/project/SendButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const handleToggleConnected = useCallback(async () => {
+    if (connected) {
+      await actions.disconnect();
+    } else {
+      await actions.connect();
+    }
+  }, [connected, actions]);
+  const handleSendToMicrobit = useCallback(async () => {
+    await actions.flash();
+  }, [actions]);
```
microbit-matt-hillsdon, line 45: This can probably just be replaced with passing actions.flash to the onClick.

src/project/SendButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+            aria-label={intl.formatMessage({ id: "more-connect-options" })}
+            data-testid="more-connect-options"
+            // Avoid animating part of the primary action change.
+            borderLeft="1px"
+            borderRadius="button"
+            as={IconButton}
+            icon={
+              <MdMoreVert
+                style={{
+                  marginLeft: "calc(-0.15 * var(--chakra-radii-button))",
```
microbit-matt-hillsdon, line 79: Is this ⋮ menu duplicated now?

src/project/project-actions.tsx
```diff
@@ -106,9 +108,10 @@ export class ProjectActions {
       await this.dialogs.show<void>((callback) => (
         <WebUSBDialog callback={callback} action={WebUSBErrorTrigger.Connect} />
       ));
+      await this.download();
     } else {
       if (await this.showConnectHelp(forceConnectHelp)) {
-        await this.connectInternal();
+        return await this.connectInternal();
```
microbit-matt-hillsdon, line 114: Do we need this to be "return await" ? Normally only needed for the scope of try/catch which we don't have here.

src/project/project-actions.tsx
```diff
@@ -454,6 +474,10 @@ export class ProjectActions {
       type: "download-main-file",
     });
 
+    if (this.isDefaultProjectName()) {
+      await this.editProjectName(true);
+    }
```
microbit-matt-hillsdon, line 479: Should this have the same cancel logic as above?

Maybe just have a really explicit method like "ensureProjectNameOrCancel" called for both?

src/project/project-actions.tsx
```diff
@@ -541,6 +565,34 @@ export class ProjectActions {
     }
   };
 
+  isDefaultProjectName = (): boolean =>
+    this.project.name === this.intl.formatMessage({ id: "untitled-project" });
```
microbit-matt-hillsdon, line 569: Another option would be checking whether `this.project.fs.name` was defined. I guess they might differ if you "OK" the project name dialog but stopping nagging you if you do that might be a good thing.

src/workbench/connect-dialogs/ConnectHelpDialog.tsx
```diff
@@ -71,61 +33,62 @@ const ConnectHelpDialogBody = () => {
...
       >
-        <Image
-          height="375px"
-          width="389px"
-          src={isMac ? connectGifMac : connectGifWin}
-          alt=""
-          border="1px solid"
-          borderColor="gray.600"
-        />
+        <Image height={375} width={418} src={selectMicrobit} alt="" />
```
microbit-matt-hillsdon, line 41: Let's discuss what makes as alt text here.

src/workbench/connect-dialogs/Overlay.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useCallback, useEffect, useState } from "react";
+import { zIndexOverlay } from "../../common/zIndex";
+import {
+  EVENT_END_USB_SELECT,
+  EVENT_START_USB_SELECT,
+} from "../../device/device";
+import { useDevice } from "../../device/device-hooks";
+
+const Overlay = () => {
+  const [selectingDevice, setSelectingDevice] = useState<boolean>(false);
```
microbit-matt-hillsdon, line 11: I'd be tempted to use chakra's useDisclosure hook.

src/project/SendButton.tsx
microbit-matt-hillsdon: If they end up being styled differently then I don't really care while we just have two. We should reconsider if we add more.

src/project/ProjectNameEditable.tsx
microbit-robert: Typescript complains because editProjectName has an optional param. I've rejigged it [`6b7b162`](https://github.com/microbit-foundation/python-editor-next/pull/763/commits/6b7b162c9c7399d766a75310e87e258351a08f55).

src/project/project-actions.tsx
microbit-robert: I think that is a better behaviour, [`2cf98b7`](https://github.com/microbit-foundation/python-editor-next/pull/763/commits/2cf98b77f736e0600c02f978367491783250b2f4).

src/project/project-actions.tsx
microbit-robert: [`2cf98b7`](https://github.com/microbit-foundation/python-editor-next/pull/763/commits/2cf98b77f736e0600c02f978367491783250b2f4).

src/project/SendButton.tsx
microbit-robert: [`4ca26cb`](https://github.com/microbit-foundation/python-editor-next/pull/763/commits/4ca26cbcae741e51178fa9ffa39485553c033390).

src/project/SendButton.tsx
microbit-robert: [`2cf98b7`](https://github.com/microbit-foundation/python-editor-next/pull/763/commits/2cf98b77f736e0600c02f978367491783250b2f4).

src/project/ProjectNameEditable.tsx
microbit-robert: Same as comment above, [`6b7b162`](https://github.com/microbit-foundation/python-editor-next/pull/763/commits/6b7b162c9c7399d766a75310e87e258351a08f55).

src/project/project-actions.tsx
```diff
@@ -97,22 +98,71 @@ export class ProjectActions {
...
         <WebUSBDialog callback={callback} action={WebUSBErrorTrigger.Connect} />
       ));
     } else {
-      if (await this.showConnectHelp(forceConnectHelp)) {
-        await this.connectInternal();
+      if (await this.showConnectCable(forceConnectHelp)) {
+        const connected = await this.connectInternal();
+        if (flash && connected) {
+          // Not sure why this is needed at the moment.
+          await new Promise((resolve) => setTimeout(resolve, 50));
```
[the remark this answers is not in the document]
microbit-matt-hillsdon: Discussed in person. We've seen a similar issue previously when writing e2e tests and I was unable to get to the bottom of it in a reasonable amount of time. I'm confident it's not something introduced with this PR. However, we may have made an error scenario much easier to reproduce.

I can trigger an error pretty reliably on 6x performance throttling. I'm trying to debug to understand. We might need a much larger delay. I've never reproduced it at human speed when clicking "Connect" then "Flash".

src/project/project-actions.tsx
```diff
@@ -541,6 +566,41 @@ export class ProjectActions {
     }
   };
 
+  isDefaultProjectName = (): boolean => this.fs.project.name === undefined;
+
+  ensureProjectName = async (): Promise<boolean> => {
+    let cancelDownload = false;
```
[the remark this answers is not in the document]
microbit-robert: Slightly, [`6899015`](https://github.com/microbit-foundation/python-editor-next/pull/763/commits/6899015fe9677435f1b2a3a400209d0808b7d9ed).

microbit-matt-hillsdon (approved):
Happy with this after we briefly discuss impact on the support guide and translation tomorrow.

PR #766, opened 2022-06-06
Remove insert code in favour of copy/paste

microbit-robert:
We've found from user testing that:
- users are frustrated that "Insert code" doesn't give adequate control over the insert location
- drag and drop is not very discoverable (perhaps because Insert code is very easily to discover)

This PR replaces "Insert code" with a slightly less discoverable copy action that allows control of insert via paste and adds a nudge to dnd on code hovers before the first drag and drop.


microbit-robert:
How long do we show copy button for?

src/editor/active-editor-hooks.tsx
```diff
@@ -14,28 +14,40 @@ import React, {
...
-      type: "code-insert",
+      type: "code-copy",
       message: id,
     });
-    this.view.dispatch(calculateChanges(this.view.state, code, type));
-    this.view.focus();
+    await navigator.clipboard.write([
+      new ClipboardItem({
+        // @ts-ignore
+        "text/html": new Blob([`<code data-type="${type}">${code}</code>`], {
```
microbit-matt-hillsdon, line 43: Presumably we should be HTML escaping the code here.
Maybe easiest to just build it as as DOM and use outerHTML?

src/editor/codemirror/dnd.ts
```diff
@@ -40,7 +41,11 @@ export type CodeInsertType =
   /**
    * A function call.
    */
-  | "call";
+  | "call"
+  /**
+   * Context is unknown, used for paste events.
+   */
+  | "unknown";
```
microbit-matt-hillsdon, line 48: Presumably we could know in a similar way to dnd? Let's discuss.

src/editor/codemirror/copypaste.ts
```diff
@@ -0,0 +1,78 @@
...
+        ?.getAttribute("data-type") as CodeInsertType) || "unknown",
+  };
+};
+
+const copyPasteHandlers = () => {
+  const textEncoder = new TextEncoder();
+  return [
+    EditorView.domEventHandlers({
+      paste(event, view) {
+        if (
```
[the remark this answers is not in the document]
microbit-matt-hillsdon: No action - note for me.

lang/en.json
```diff
@@ -479,6 +491,10 @@
     "defaultMessage": "You must <link>update your firmware before</link> you can connect to this micro:bit.",
     "description": "Text in the no micro:bit found dialog"
   },
+  "open-action": {
```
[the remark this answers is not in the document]
microbit-robert: There's now a load-action which is "open..." and an open-action which "open". We can merge them if we decide what we want to do with the ellipses.

src/editor/codemirror/dnd.ts
microbit-matt-hillsdon: Does the unknown case happen anymore? It looks like the data attribute takes care of it.

If we systematically set the data attribute then we could skip handling anything without it (assuming it's just other random HTML on the clipboard).

microbit-robert: I've just removed it. If we somehow end up with html/text on the clipboard without an insert type, it pastes just fine, as it would for an "example".

src/documentation/api/ApiNode.tsx
```diff
@@ -385,74 +393,95 @@ const DraggableSignature = ({
...
+          </Text>
+        </HStack>
+      </Tooltip>
+      <CodeActionButton
+        isOpen={copyCodeButton.isOpen}
+        toHighlighted={highlight.onOpen}
+        toDefault={highlight.onClose}
+        codeAction={handleCopyCode}
+        borderAdjustment={false}
+        toolkitType={"api"}
```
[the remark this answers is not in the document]
microbit-robert: [`62e7bd9`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/62e7bd970ea6fd32476da54b503af5597a5b65c5)

src/editor/active-editor-hooks.tsx
microbit-robert: [`5eac9df`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/5eac9df004ad23d739064f5f22be3e9e3791e90c).

src/editor/codemirror/copypaste.ts
```diff
@@ -0,0 +1,78 @@
...
+        deployment.logging.event({
+          type: "paste",
+          value: lineCount,
+        });
+
+        const lineNumber = view.state.doc.lineAt(
+          view.state.selection.ranges[0].from
+        ).number;
+
+        view.dispatch(
```
microbit-matt-hillsdon, line 61: Is it possible to tag this transaction as "input.paste" so that the normal line count code runs? 

Or perhaps we shouldn't even run it, as it's not desirable to have an event that conflates regular pastes and reference/api pastes. Let's discuss this.

src/editor/codemirror/dnd.ts
microbit-robert: [`4eaa87b`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/4eaa87bfc6ceea88e24cb03ebca2a89fa9cbea7c).

src/editor/codemirror/copypaste.ts
```diff
@@ -0,0 +1,78 @@
...
+  type: CodeInsertType;
+}
+
+const getCodeFromHtml = (
+  innerHTML: string | undefined
+): PasteContext | undefined => {
+  if (!innerHTML) {
+    return;
+  }
+  const parent: HTMLElement = document.createElement("html");
```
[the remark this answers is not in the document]
microbit-robert: [`5fcf5a2`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/5fcf5a20adef464784f96400eb5d6c2d7739dcda).

src/editor/codemirror/copypaste.ts
```diff
@@ -0,0 +1,78 @@
...
+  return [
+    EditorView.domEventHandlers({
+      paste(event, view) {
+        if (
+          !view.state.facet(EditorView.editable) ||
+          event.clipboardData?.getData("text")
+        ) {
+          return;
+        }
+        event.preventDefault();
```
microbit-matt-hillsdon, line 39: I feel like we shouldn't call this unless we have text to insert. Otherwise we might prevent some CM logic. Probably not, given the earlier tests, but I think structuring like that makes sense. So, "is it the weird special case" -> e.preventDefault() and do the action. Otherwise we just return and CM runs.

src/editor/codemirror/copypaste.ts
microbit-robert: It might be better to log it in the same we do for dnd drop events.

microbit-matt-hillsdon:
@microbit-robert, should the API entries also have the drag hint tooltip? They look the same so having one flag shared between them is probably fine. If this is involved then as API is less of a focus we can split it out.

microbit-robert:
> @microbit-robert, should the API entries also have the drag hint tooltip? They look the same so having one flag shared between them is probably fine. If this is involved then as API is less of a focus we can split it out.

The API entries should already have the tooltip, and the flag is currently shared between them, but it would be straight forward to change this to two separate flags.

src/App.tsx
```diff
@@ -64,10 +69,17 @@ const App = () => {
...
 
-  const settings = useLocalStorage<Settings>(
+  const settings = useStorage<Settings>(
+    "local",
     "settings",
-    isValidSettingsObject,
-    defaultSettings
+    defaultSettings,
+    isValidSettingsObject
+  );
```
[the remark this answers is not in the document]
microbit-robert: [`904dc76`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/904dc76da1572d0b20d895766037abf806bca1d1).

microbit-matt-hillsdon:
> > @microbit-robert, should the API entries also have the drag hint tooltip? They look the same so having one flag shared between them is probably fine. If this is involved then as API is less of a focus we can split it out.
> 
> The API entries should already have the tooltip, and the flag is currently shared between them, but it would be straight forward to change this to two separate flags.

Ah, it is refresh preserving session storage that confused me into thinking this wasn't working. All good, I think the current shared flag makes sense.

lang/en.json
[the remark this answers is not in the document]
microbit-matt-hillsdon: Let's change the ids. That'll give an opportunity to retranslate Load too which may now be wrong.

src/editor/codemirror/dnd.ts
microbit-matt-hillsdon: When reading from the clipboard let's say there's no "code" if there's no data-attribute. We always set it and we're best off not trying to paste HTML from elsewhere.

src/editor/codemirror/copypaste.ts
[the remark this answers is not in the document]
microbit-matt-hillsdon: Thanks, looks good.

src/editor/codemirror/copypaste.ts
microbit-matt-hillsdon: Discussed. Let's log it analogously to dnd and not log it as a user paste at all.

src/editor/codemirror/copypaste.ts
microbit-matt-hillsdon: This change is fine given the plan on https://github.com/microbit-foundation/python-editor-next/pull/766#discussion_r898917406.

src/editor/codemirror/copypaste.ts
microbit-robert: [`738ca6a`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/738ca6a4205e75ce2349f6b8ace0eeb6c333c4a2). A user paste is logged as `paste` with number of lines, and, as with all of the logging events that handle code examples, the special code paste is logged as `code-paste` with the code id.

lang/en.json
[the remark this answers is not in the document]
microbit-robert: [`4d823de`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/4d823de5865eddd66090f76739edf2a22280faba).

src/editor/codemirror/dnd.ts
microbit-robert: [`6d4d186`](https://github.com/microbit-foundation/python-editor-next/pull/766/commits/6d4d186273f2ad7499dffee1509852d87ddff711).

src/e2e/app.ts
```diff
@@ -746,6 +781,55 @@ export class App {
     this.page = this.createPage();
     page = await this.page;
     await page.goto(this.url);
+
+    // Install fake clipboard due to problems with pupeteer's native clipboad support.
```
microbit-matt-hillsdon, line 785:
```suggestion
    // Install fake clipboard due to problems with puppeteer's native clipboard support.
```

microbit-matt-hillsdon (approved):
Looking good now.

PR #769, opened 2022-06-07
Prefer e.g. random over urandom.

microbit-matt-hillsdon:
Add missing gc, sys, struct modules that we already have docs for
(spotted during translation prep).

Other missing modules are missing docs and won't be addressed for live,
or are Easter-egg-type modules and intentionally omitted.


microbit-robert (approved):
LGTM.

Issue #531, opened 2022-02-02
Lowercase restriction for filenames prevents use of characters from some languages 


microbit-matt-hillsdon:
Continuing from #530, the lowercase restriction isn't quite right either, e.g. 共和国 .py works fine but we reject it.

There are three sources of constraints:

1. What can be created as a file in MicroPython's file system
1. What can be referenced via an import in the grammar
1. Any best practice constraints documented for Python that we want to apply

We should review these and amend the validation, referencing sources where we can. I can't recall the original source for the constraints.

PR #770, opened 2022-06-14
Fix module name validation.

microbit-matt-hillsdon:
We haven't checked the exact unicode ranges as that's quite involved but
this should generally be more accurate and no longer ban numbers or
characters from languages without an upper/lowercase distinction.

Closes https://github.com/microbit-foundation/python-editor-next/issues/524
Closes https://github.com/microbit-foundation/python-editor-next/issues/531

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -27,10 +27,6 @@
     "defaultMessage": "Added file {changeName}",
     "description": "Change made to file"
   },
-  "already-exists": {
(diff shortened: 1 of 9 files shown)


microbit-robert:
This all looks good. One query on the UI. You can end up with a warning on the way to adding the `.py` extension to the file name. 

image
image

microbit-robert (approved):
LGTM.

Issue #761, opened 2022-05-30
`Projects` -> `Add file` might be interpreted as "upload/load file"


microbit-carlos:
If a user wanted to "load" a local file (like a MicroPython accessory module) to Editor they have to go to the "Projects" section and the two buttons presented are:

<img width="478" alt="image" src="https://user-images.githubusercontent.com/29712657/170967778-6942c0b4-2920-4339-8804-84f949251b08.png">

I think some people might read the `Add file` button as "Add a file from your computer". With that in mind I think something like `Add new file` or `Create file`/`Create new file` might be a bit more clear.

microbit-matt-hillsdon:
Personally I find both "Create file" or "Create new file" clearer. However, we changed it from "Create new file" to avoid the misconception seen in the first round of user testing that users were creating a new project. Before changing it back (more or less) we should consider this scenario. Does the Projects tab give sufficient context to prevent the confusion?

The original change was made on #664 along with Files -> Project and using the project name as the Project tab header.

microbit-carlos:
Yeah, I could see how `Create new file` could maybe misinterpreted by somebody looking for a way to create a new project.

In that case I think maybe `Create file` hits a better middle ground, without the word "new" is less likely to be interpreted as "new project".

microbit-matt-hillsdon:
I had a support interaction today where "Add file" was misunderstood in this way. I think we should go ahead and make this change (and soon, so it's in place for translation).

PR #771, opened 2022-06-14
"Add file" -> "Create file"

microbit-matt-hillsdon:
Update related text.

Closes https://github.com/microbit-foundation/python-editor-next/issues/761

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -23,25 +23,9 @@
     "defaultMessage": "Read more about MicroPython",
     "description": "Arial label to expand a collapsed section about MicroPython"
   },
-  "add-action": {
-    "defaultMessage": "Add",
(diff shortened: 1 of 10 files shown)


microbit-robert:
@microbit-matt-hillsdon Should we also change the `addFile` function in project-actions to `createFile`. There is also a logging event in this function which is `add-file` but maybe that's better left alone from an analytics point of view.

microbit-matt-hillsdon:
> @microbit-matt-hillsdon Should we also change the `addFile` function in project-actions to `createFile`. There is also a logging event in this function which is `add-file` but maybe that's better left alone from an analytics point of view.

Changed both.

microbit-robert (approved):
LGTM.

PR #772, opened 2022-06-20
Add ability to collapse/expand sidebar.

microbit-robert:


src/workbench/SideBarTab.tsx
```diff
@@ -21,10 +23,17 @@ const SideBarTab = ({
   mb,
   handleTabClick,
   active,
+  tabIndex,
 }: SideBarTabProps) => {
   const width = "5rem";
+  const ref = useRef<HTMLButtonElement>(null);
+  useEffect(() => {
+    // Override tabindex.
+    ref.current!.setAttribute("tabindex", "0");
```
microbit-matt-hillsdon, line 32: Why does this need to be an effect?

microbit-robert:
We can revisit the window width at which we load the Editor with the sidebar closed after further improvements have been made to maximise the use of available space.

PR #778, opened 2022-06-23
Download hex file by default when WebUSB not available.

microbit-robert:


microbit-robert:
This was in the spec for changing the flash button, but we accidentally changed the behaviour before merge. In the beta, clicking connect when WebUSB is not available will already download the hex.

microbit-matt-hillsdon:
Thanks. @microbit-robert please can you drop Lucy a note to explain the change as she'll be revisiting the dialog?

PR #779, opened 2022-06-23
Update package.json file.

microbit-robert:
Removes dependencies on `react-copy-to-clipboard` which is no longer used.


microbit-matt-hillsdon:
Just spotted this locally 😄

PR #791, opened 2022-06-27
Improve connect help dialogs, add hex transfer help dialog.

microbit-robert:
Simplifies the `Not Found` dialog and includes a `download` action for the user to easily download their hex after failing to connect to their micro:bit. Adds a new post-download dialog to help the user transfer the downloaded hex file to their device via drag and drop.


microbit-robert:
In draft, as e2e tests will likely need altering and expanding.

src/workbench/connect-dialogs/TransferHexDialog.tsx
```diff
@@ -0,0 +1,145 @@
...
+            }}
+          />
+        </Text>
+      </VStack>
+      <Flex justifyContent="center" width="100%">
+        {isMac ? (
+          <Image width="100%" height="100%" src={transferHexMac} alt="" />
+        ) : (
+          <Image width="100%" height="100%" src={transferHexWin} alt="" />
+        )}
```
microbit-matt-hillsdon, line 109: Might be nicer to determine the src with a branch and only have one Image tag?

I think no alt is OK given that the sentence above is explaining the drag.

microbit-matt-hillsdon (approved):
Looks good. One comment above but no strong view.

I think I might even be coming around to the bold so I'm not going to mention it.

src/workbench/connect-dialogs/TransferHexDialog.tsx
microbit-robert: Yes, that's sensible. This was leftover from when width and height values were different for each. [`3f05893`](https://github.com/microbit-foundation/python-editor-next/pull/791/commits/3f05893beb09ad020a91f341d50800055cdbc313).

PR #797, opened 2022-07-04
Rename Download -> Save.

microbit-matt-hillsdon:
- We want ‘Send to micro:bit’ to be the preferred route for flashing -
  if you can’t use WebUSB it downloads and gives you instructions anyway
  and if it fails then the download route is now a clear alternative

- We know that some people aren’t sure how to save to backup their work
  and don’t realise that download does the same thing. Save is the label
  more folk are familiar with.

- Long term if we have other ways to save (say cloud storage) then Save
  is the better label to hang these off.

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -623,15 +623,15 @@ export class App {
   }
 
   /**
-   * Trigger a download but don't wait for it to complete.
+   * Trigger a save but don't wait for it to complete.
    *
    * Useful when the action is expected to fail.
-   * Otherwise see waitForDownload.
+   * Otherwise see waitForSave.
    */
-  async download(): Promise<void> {
+  async save(): Promise<void> {
     const document = await this.document();
-    const downloadButton = await document.getByText("Download");
-    return downloadButton.click();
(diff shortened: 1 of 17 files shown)


microbit-matt-hillsdon:
This PR requires an upgrade of the theme package before merge.

src/e2e/app.ts
```diff
@@ -796,12 +796,12 @@ export class App {
   }
 
   /**
-   * Trigger a download and wait for it to complete.
+   * Trigger a hex ifle save and wait for the download to complete.
```
microbit-robert, line 799:
```suggestion
   * Trigger a hex file save and wait for the download to complete.
```

microbit-matt-hillsdon: `ifle` feels like it could be a Python keyword 😄

microbit-robert (approved):
LGTM.

microbit-matt-hillsdon:
> This PR requires an upgrade of the theme package before merge.

Done (and rebased).

microbit-matt-hillsdon:
Will merge when CI passes.

PR #799, opened 2022-07-05
Minimal while smooshing

microbit-matt-hillsdon:
"While smooshing" avoids the nested `while True` constructs that are
easy to create from the simple examples available to drag and drop.

This doesn't cope with while loops with a preamble (see e.g. variables
section). For that we'd need to split the preamble which would add a
fair amount of complexity. It copes well with the examples in Buttons
and Accelerometer though. We may revisit this in future.

We also considered highlighting the `while True` in yellow and may revisit
that too.

### DIFF
--- a/src/editor/codemirror/edits.test.ts
+++ b/src/editor/codemirror/edits.test.ts
@@ -215,6 +215,24 @@ describe("edits", () => {
       expected: "import radio\n\n\npass\nradio.off()\n",
     });
   });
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
Needs tests for the new indents module and the new behaviour in edits.ts.
More importantly though we need to decide if the behaviour is sufficiently non-spooky.

microbit-robert (approved):
LGTM.

PR #807, opened 2022-07-08
Reduce spacing around top level reference items.

microbit-robert:


microbit-robert:
Using the built-in breakpoints is a bit messy as it's mobile first, so the arrays are full of repeat values as we only use one specific breakpoint so far. The built-in breakpoints are based on width. We could use the `useMediaQuery` hook instead and keep the queries in a separate file to keep track (similar to z-index).

src/common/AreaHeading.tsx
```diff
@@ -4,18 +4,19 @@
  * SPDX-License-Identifier: MIT
  */
 import { Box, Text } from "@chakra-ui/layout";
+import { ReactNode } from "react";
 
 interface AreaHeadingProps {
   name: string;
-  description: string;
+  description: string | ReactNode;
```
microbit-matt-hillsdon, line 11: I think this can just be ReactNode.

microbit-robert: [`9a2a136`](https://github.com/microbit-foundation/python-editor-next/pull/807/commits/9a2a13661160c3d4201c90b1113b0dbf50cf5934).

PR #811, opened 2022-07-12
Flag to choose alt MicroPython with SoundEffect.

microbit-matt-hillsdon:
We also need matching stubs so reorg to tie these closer together.

Demos:

- https://review-python-editor-next.microbit.org/audio-sound-effect/?flag=audioSoundEffect
- https://review-python-editor-next.microbit.org/audio-sound-effect/?flag=none (only necessary as review branches auto-enable flags)

Perhaps in the future we could expose a setting for this rather than the flag.


src/language-server/client.ts
```diff
@@ -170,12 +171,9 @@ export class LanguageServerClient extends EventEmitter {
   }
 
   private async getInitializationOptions(): Promise<any> {
+    const branch = microPythonConfig.stubs;
     const typeshed = await retryAsyncLoad(() => {
-      switch (this.locale) {
```
microbit-matt-hillsdon, line 174: Need to review whether this is still bundle splitting correctly. I think it's cleverer than I give it credit for though.

microbit-matt-hillsdon:
Not a draft but relying on @microbit-robert to test on a device.

microbit-robert (approved):
LGTM. Tested and working on device. Flag is effective in controlling MicroPython version and stubs.

PR #814, opened 2022-07-13
Custom error message for loading incompatible MakeCode hex.

### LINKED ISSUE #617: Better message when opening hex file from MakeCode.
V2 behaviour:

image

V3 behaviour:
image

Neither are amazing but the V3 one is worse.

microbit-robert:
Closes #617.


microbit-robert:
We should find out if there is a more definitive way to determine whether a hex file was created with MakeCode.

PR #815, opened 2022-07-14
Add overlay for files marked as modules to prevent unintentional editing.

### LINKED ISSUE #80: Add friction to editing modules with magic marker to avoid accidents
We should consider adding some friction to editing modules with magic marker to avoid accidental changes.

E.g. the top bar could show an "Edit" button.

microbit-robert:
The intent is to avoid user edits to modules supplied by accessory vendors and similar, who use a special comment to tag their files as modules. Provide a setting to allow the curious (or the vendors themselves) to make edits.

We also considered:
- A single click to edit a particular file, but Jonny preferred the increased friction of the trip via Settings
- Jonny also proposed as an alternative a "type the name of this module" dialog or embedded form to enable editing but this seemed too much friction to me.

Closes #80

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -655,6 +655,10 @@
     "defaultMessage": "Serial terminal",
     "description": "Aria label for the serial terminal"
   },
+  "setting-allow-editing-third-party": {
+    "defaultMessage": "Allow editing third-party modules",
+    "description": "Checkbox setting label"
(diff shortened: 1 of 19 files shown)


microbit-robert:
Todo:
- More inspired UI.
- Input on text so UI strings can be created/used.

PR #820, opened 2022-07-15
Add post-save dialog and update flow.

microbit-robert:
If WebUSB is supported, saving a hex will trigger a post-save dialog, providing information about the hex downloaded. It also offers a link to the transfer hex help dialog.

On saving the main script from the additional options on the save button, the main script is downloaded, and is now suffixed with "main.py". If there are multiple files in the project, a dialog is shown to warn the user that only main.py has been saved, and advises on how to download other the Python files in the project.

On flashing/connecting when WebUSB is not supported, the WebUSB dialog is shown once per session, followed by the transfer hex help dialog, there after, only the transfer hex help dialog is shown.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -515,6 +515,26 @@
     "defaultMessage": "Permanently delete {filename}?",
     "description": "Confirmation question to permanently delete file"
   },
+  "post-save-message-files": {
+    "defaultMessage": "This contains all files in this project (including any additional files, e.g. to run accessories, listed in the Project section).",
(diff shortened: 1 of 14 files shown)


microbit-matt-hillsdon:
Code-wise this looks fine.

There's an e2e update to do for the new dialog flow after save.

@microbit-robert, please can we also share the current dialogs for discussion in chat? I'll leave that until you are back.

microbit-matt-hillsdon (approved):
Looks good.

PR #822, opened 2022-07-15
Fix missing string when project name is blank.

microbit-matt-hillsdon:
Fix consistency with the Create file validation phrasing.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -275,10 +275,6 @@
     "defaultMessage": "The name should be all lowercase",
     "description": "Warning text for new Python file with uppercase"
   },
-  "file-name-not-blank": {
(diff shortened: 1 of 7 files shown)


lang/en.json
```diff
@@ -275,10 +275,6 @@
     "defaultMessage": "The name should be all lowercase",
     "description": "Warning text for new Python file with uppercase"
   },
-  "file-name-not-blank": {
```
microbit-matt-hillsdon, line 278: This was a project name specific string but I think was unintentionally reprefixed along with the Create file related strings in the rejig of the validation for that dialog.

microbit-robert (approved):
LGTM.

PR #828, opened 2022-07-20
Improve multiple files dialog title.

microbit-robert:
Also adds links to plural syntax format used to each UI string description where relevant.

### DIFF
--- a/src/common/MultipleFilesDialog.tsx
+++ b/src/common/MultipleFilesDialog.tsx
@@ -4,10 +4,10 @@
  * SPDX-License-Identifier: MIT
  */
 import { Text, VStack } from "@chakra-ui/react";
-import { ReactNode, useEffect, useState } from "react";
+import { ReactNode } from "react";
 import { FormattedMessage } from "react-intl";
 import { GenericDialog, GenericDialogFooter } from "../common/GenericDialog";
-import { useFileSystem } from "../fs/fs-hooks";
+import { useProject } from "../project/project-hooks";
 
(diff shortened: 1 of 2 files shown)


src/common/MultipleFilesDialog.tsx
```diff
@@ -37,6 +38,15 @@ export const MultipleFilesDialog = ({ callback }: MultipleFilesDialogProps) => {
...
 const MultipleFilesDialogBody = () => {
+  const fs = useFileSystem();
+  const [fileCount, setFileCount] = useState<number>(1);
+  useEffect(() => {
+    const getNumFiles = async () => {
+      const result = (await fs.statistics()).files;
+      setFileCount(result);
+    };
+    getNumFiles();
+  }, [fs, setFileCount]);
```
microbit-matt-hillsdon, line 49: I don't see how this would update as required. It kinda works but only because the dialog is a modal so it works to fetch it once but it seems a fragile pattern (fs and setFileCount will never change identity).

Please use the useProject() hook and project.files.length.

microbit-robert: Fixed here and where the same pattern was used elsewhere - [`e958238`](https://github.com/microbit-foundation/python-editor-next/pull/828/commits/e958238a62f19c96cd967a0e73d023b27a0f35a9).

microbit-matt-hillsdon: Thanks.

microbit-matt-hillsdon (approved):
LGTM.

Issue #827, opened 2022-07-20
Flags need to opt-in to being enabled for review branches


microbit-matt-hillsdon:
Flags need to opt-in to being enabled for review branches as some more recent flags (e.g. to use an alternative MicroPython) are too confusing to enable by default.

PR #831, opened 2022-07-21
Change flags to opt in to being enabled for stages.

microbit-matt-hillsdon:
It's no longer appropriate to enable all flags by default for REVIEW
stage.

Closes https://github.com/microbit-foundation/python-editor-next/issues/827

### DIFF
--- a/src/common/sanity.ts
+++ b/src/common/sanity.ts
@@ -58,7 +58,7 @@ export const fetchContent = async <T>(
 };
 
 const fetchContentInternal = async (query: string): Promise<any> => {
-  const response = await fetch(queryUrl(query));
+  const response = await fetch(queryUrl(query), { credentials: "include" });
   if (response.ok) {
     const { result } = await response.json();
     if (!result) {
--- a/src/common/sanity.ts
+++ b/src/common/sanity.ts
@@ -58,7 +58,7 @@ export const fetchContent = async <T>(
 };
 
 const fetchContentInternal = async (query: string): Promise<any> => {
-  const response = await fetch(queryUrl(query), { credentials: "include" });
+  const response = await fetch(queryUrl(query));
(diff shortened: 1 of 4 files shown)


src/common/sanity.ts
```diff
@@ -58,7 +58,7 @@ export const fetchContent = async <T>(
 };
 
 const fetchContentInternal = async (query: string): Promise<any> => {
-  const response = await fetch(queryUrl(query));
+  const response = await fetch(queryUrl(query), { credentials: "include" });
```
microbit-matt-hillsdon, line 61: Oops, unrelated change.

```suggestion
  const response = await fetch(queryUrl(query));
```

src/documentation/ideas/content.ts
```diff
@@ -14,7 +14,7 @@ const ideasQuery = (languageId: string): string => {
     throw new Error("Invalid language id.");
   }
   return `
-    *[_type == "pythonIdea" && language == "${languageId}" && !(_id in path("drafts.**"))]{
+    *[_type == "pythonIdea" && language == "${languageId}" ]{
```
microbit-matt-hillsdon, line 17:
```suggestion
    *[_type == "pythonIdea" && language == "${languageId}" && !(_id in path("drafts.**"))]{
```

microbit-robert (changes requested):
Please see comment above.

microbit-matt-hillsdon:
> Using `flag=none` does not seem to disable flags enabled by stage. Using `flag=*` seems to disable all flags rather than enabling them.

Added tests and fix in 465041ab90a7229b099e38d8134ae4c1394d3d00. The logic is also easier to follow now.

microbit-robert (approved):
Thanks. LGTM.

PR #836, opened 2022-07-21
Fix docs heading size.

microbit-matt-hillsdon:
Chakra's Heading is responsive in a way I didn't anticipate.
Switch to Text though still with the correct markup.


microbit-robert (approved):
Looks good. I think the font-weight for the Reference h3 is semi-bold at 600. Bold is 700, but we don't have that weight so it still renders at 600. Might be worth tweaking if this is the case.

microbit-matt-hillsdon:
Yeah ok, I should probably use semibold. It's frustrating that we have both but they're the same. Ideally we'd fix it in the theme.

microbit-matt-hillsdon:
Still looks good on PR preview.

PR #837, opened 2022-07-21
Update ideas query to retrieve ideas in config order.

microbit-robert:


src/documentation/ideas/content.ts
```diff
@@ -14,25 +14,29 @@ const ideasQuery = (languageId: string): string => {
...
     }`;
 };
 
 const adaptContent = (result: any): Idea[] | undefined => {
-  const ideas = result as Idea[];
-  if (ideas.length === 0) {
-    return undefined;
+  const ideas = result?.pythonIdeasOrder as Idea[];
+  if (ideas) {
+    if (ideas.length === 0) {
```
microbit-matt-hillsdon, line 37: This looks slightly odd.

microbit-matt-hillsdon:
Confirmed fr to en fallback is working. Some French ideas will be lost until the next translation sync though we're planning to run an extra one today.

microbit-matt-hillsdon (approved):
Query looks good. One minor comment above.

src/documentation/ideas/content.ts
microbit-robert: More logical check here -> [`80ad3b2`](https://github.com/microbit-foundation/python-editor-next/pull/837/commits/80ad3b25970c049196b391afded3b030e5e20292).

PR #841, opened 2022-07-22
Reset project

### LINKED ISSUE #840: File system dirty check breaks after restore from session storage
Repro:

1. New tab
2. Edit
3. Refresh, OK warning
4. Refresh, note no warning

The dirty state is in-memory only.

This will become important as we're planning to use the same dirty state to avoid showing superfluous warning dialogs when loading files.

microbit-matt-hillsdon:
- Rearrange Project tab to support a new "Reset project" button
- Reset button resets to default starter code
- Skip the scary Replace dialogs when the file system is not dirty (initially, and after Save and main.py download - if single file). Was affected by #840 which is now fixed by this PR.

Closes #840.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -147,6 +147,10 @@
     "defaultMessage": "Confirm and save",
     "description": "Confirm and save action label"
   },
+  "confirm-save-hint": {
+    "defaultMessage": "Cancel then use <strong>Save</strong> to keep a copy of your project.",
+    "description": "Hint shown when the user has unsaved changes that they might lose if they continue with an action"
+  },
   "connect-action": {
     "defaultMessage": "Connect",
     "description": "Connect button text"
--- a/lang/en.json
+++ b/lang/en.json
@@ -148,7 +148,7 @@
     "description": "Confirm and save action label"
   },
   "confirm-save-hint": {
-    "defaultMessage": "Cancel then use <strong>Save</strong> to keep a copy of your project.",
+    "defaultMessage": "Cancel then use Save to keep a copy of your project.",
     "description": "Hint shown when the user has unsaved changes that they might lose if they continue with an action"
   },
   "connect-action": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -132,7 +132,7 @@
     "description": "Confirmation message body for replacing project dialog"
   },
   "confirm-replace-reset": {
-    "defaultMessage": "Replace all files with the initial project?",
+    "defaultMessage": "Replace all files with the default project?",
     "description": "Confirmation message for the reset action"
   },
   "confirm-replace-title": {
@@ -616,7 +616,7 @@
     "description": "Action to reset the project to its default state"
   },
   "reset-project-hover": {
-    "defaultMessage": "Resets the project to the initial code, discarding your work",
+    "defaultMessage": "Resets the project to the default code, discarding your work",
     "description": "Reset action hover text"
   },
   "results-count": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -615,6 +615,10 @@
     "defaultMessage": "Reset project",
     "description": "Action to reset the project to its default state"
   },
+  "reset-project-feedback": {
+    "defaultMessage": "Project reset to the default code",
+    "description": "Confirmation message after resetting the project"
+  },
   "reset-project-hover": {
     "defaultMessage": "Resets the project to the default code, discarding your work",
     "description": "Reset action hover text"
--- a/lang/en.json
+++ b/lang/en.json
@@ -616,11 +616,11 @@
     "description": "Action to reset the project to its default state"
   },
   "reset-project-feedback": {
-    "defaultMessage": "Project reset to the default code",
+    "defaultMessage": "Project reset to the default starter code",
     "description": "Confirmation message after resetting the project"
   },
   "reset-project-hover": {
-    "defaultMessage": "Resets the project to the default code, discarding your work",
+    "defaultMessage": "Resets the project to the default starter code, discarding your work",
     "description": "Reset action hover text"
   },
   "results-count": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -131,6 +131,10 @@
     "defaultMessage": "Replace all files with those in the hex?",
     "description": "Confirmation message body for replacing project dialog"
   },
+  "confirm-replace-reset": {
+    "defaultMessage": "Replace all files with the initial project?",
+    "description": "Confirmation message for the reset action"
+  },
   "confirm-replace-title": {
     "defaultMessage": "Confirm replace project",
     "description": "Confirmation message title for replacing project dialog"
@@ -606,6 +610,14 @@
   "replace-action-label": {
     "defaultMessage": "Replace",
     "description": "Action label for replacing project dialog"
+  },
+  "reset-project-action": {
+    "defaultMessage": "Reset project",
+    "description": "Action to reset the project to its default state"
+  },
+  "reset-project-hover": {
+    "defaultMessage": "Resets the project to the initial code, discarding your work",
+    "description": "Reset action hover text"
   },
   "results-count": {
     "defaultMessage": "{count, plural, =0 {No results} one {# result} other {# results}}",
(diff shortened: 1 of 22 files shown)


lang/en.json
```diff
@@ -131,6 +131,10 @@
     "defaultMessage": "Replace all files with those in the hex?",
     "description": "Confirmation message body for replacing project dialog"
   },
+  "confirm-replace-reset": {
+    "defaultMessage": "Replace all files with the default project?",
```
microbit-matt-hillsdon, line 135: This should be "default starter code". I thought I'd changed this.

src/project/project-actions.tsx
```diff
@@ -543,6 +596,10 @@ export class ProjectActions {
             showMultipleFilesHelp: false,
           });
         }
+        if (this.project.files.length === 1) {
+          // Saving the main file is an OK way to reset the dirty flag if there are no other files.
+          await this.fs.clearDirtyFlag();
+        }
```
microbit-matt-hillsdon, line 602: Rob points out there's some risk that the user won't actually have the file as browsers default to suggesting you discard .py file downloads. Perhaps we should just drop this bit.

lang/en.json
microbit-matt-hillsdon: dedf764fc5300b8b6ceb61094a00d133e9f3d4ba

src/project/project-actions.tsx
microbit-matt-hillsdon: I don't have a strong view on this.

microbit-matt-hillsdon:
Still needs e2e tests for when you are and aren't prompted.

src/project/project-actions.tsx
```diff
@@ -325,12 +335,54 @@ export class ProjectActions {
...
+    });
+    await this.openProject(
+      defaultInitialProject,
+      this.intl.formatMessage({
+        id: "confirm-replace-reset",
+      })
+    );
+    this.actionFeedback.success({
+      title: this.intl.formatMessage({ id: "reset-project-feedback" }),
+    });
```
microbit-robert, line 383: This currently runs even if you clicked "Cancel"

src/project/project-actions.tsx
```diff
@@ -325,12 +335,54 @@ export class ProjectActions {
...
+      { id: "confirm-replace-with-idea" },
+      { ideaName: pythonProject.projectName }
+    );
+    await this.openProject(pythonProject, confirmPrompt);
+    this.actionFeedback.success({
+      title: this.intl.formatMessage(
+        { id: "loaded-file-feedback" },
+        { filename: title }
+      ),
+    });
```
microbit-robert, line 368: This also runs if you click "Cancel" after opening an idea.

src/project/ProjectArea.tsx
```diff
@@ -24,68 +23,68 @@ const ProjectArea = ({
...
-                onSelectedFileChanged(f.name);
-              }
-            };
-            return (
-              <ListItem
-                key={f.name}
-                fontWeight={selected ? "semibold" : undefined}
-                _hover={{
-                  bgColor: "blimpTeal.100",
+    <VStack spacing={5} pt={2} flex="1 0 auto" height={0} alignItems="stretch">
```
microbit-robert, line 29: The spacing looks a little generous between the top of the files list and the editable project name in my opinion.

microbit-matt-hillsdon: Thanks, I'll take a look. I think this it fell out this way from rearranging components.

src/project/ProjectArea.tsx
```diff
@@ -24,68 +23,68 @@ const ProjectArea = ({
   selectedFile,
   onSelectedFileChanged,
 }: ProjectAreaProps) => {
-  const { files, name: projectName } = useProject();
+  let { files, name: projectName } = useProject();
+  //files = Array(20).fill(files[0]);
```
microbit-robert, line 27: Should revert once no longer needed for testing scrollable file list area.

src/fs/fs.ts
```diff
@@ -249,10 +250,15 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
    */
   async setProjectName(projectName: string) {
     await this.storage.setProjectName(projectName);
-    this._dirty = true;
+    this.markDirty();
```
microbit-robert, line 253:
```suggestion
    await this.markDirty();
```

src/fs/fs.ts
```diff
@@ -428,6 +432,11 @@ export class FileSystem extends EventEmitter implements FlashDataSource {
     return fs.getUniversalHex();
   }
 
+  async clearDirty(): Promise<void> {
+    this._dirty = false;
+    this.storage.clearDirty();
```
microbit-robert, line 437:
```suggestion
    return this.storage.clearDirty();
```

src/project/project-actions.tsx
microbit-matt-hillsdon: 493724d287a5f984b1dacc52ea7aa92fd0e0d839

src/project/project-actions.tsx
microbit-matt-hillsdon: 493724d287a5f984b1dacc52ea7aa92fd0e0d839

src/fs/storage.ts
```diff
@@ -43,7 +51,7 @@ export class InMemoryFSStorage implements FSStorage {
     return this._data.has(filename);
   }
 
-  async setProjectName(projectName: string) {
+  async setProjectName(projectName: string | string) {
```
microbit-robert, line 54:
```suggestion
  async setProjectName(projectName: string | undefined) {
```

src/project/ProjectArea.tsx
microbit-matt-hillsdon: Halved in b46219c64b4c0cf34766bbc04bf537f42c38e434

src/project/ProjectArea.tsx
microbit-matt-hillsdon: Done in b46219c64b4c0cf34766bbc04bf537f42c38e434

src/fs/fs.ts
microbit-matt-hillsdon: Absolutely, thanks,

src/fs/storage.ts
microbit-matt-hillsdon: 😂

src/fs/storage.ts
```diff
@@ -72,10 +80,21 @@ export class InMemoryFSStorage implements FSStorage {
...
+    return this._dirty;
+  }
 }
 
-const fsPrefix = "fs/";
-const projectNameKey = "projectName";
+const fsFilesPrefix = "fs/files/";
+const fsMetadataPrefix = "fs/meta/";
+const projectNameKey = fsMetadataPrefix + "projectName";
+const dirtyKey = fsMetadataPrefix + "dirty";
```
microbit-matt-hillsdon, line 97: As discussed with Rob, we acknowledge this will break restore from session storage in the beta for users unlucky enough to rely on it for a page refresh before and after this PR going in.

src/project/project-actions.tsx
microbit-robert: This currently doesn't work as it needs moving out of it's immediate parent block. I find it strange that it asks for confirmation after saving main.py, so my preference is keeping this once fixed.

microbit-matt-hillsdon: Ooh, I meant to writ e2e for this and the regular Save reset but didn't.

microbit-matt-hillsdon: Scope fixed in a121783f2c6cc31258fb3779d5c6a245684c85e3

microbit-matt-hillsdon: e2e tests added too.

src/project/ProjectAreaNav.tsx
```diff
@@ -3,21 +3,21 @@
...
-      {...props}
-    >
-      <NewButton mode="button" my={1} mx={1} />
-      <OpenButton mode="button" my={1} mx={1} />
+    <Flex {...props} direction="column" alignItems="center" p={5} pb={6}>
+      <VStack alignItems="stretch" spacing={3}>
+        <NewButton mode="button" />
+        <OpenButton mode="button" />
+        <Box>
+          <ResetButton mode="button" mt={5} color="#ae1f1f" />
```
microbit-matt-hillsdon, line 18: We'll look again at the colorScheme issue here next week but won't block on it.

PR #846, opened 2022-07-22
Python Editor -> Python editor (case change only).

microbit-matt-hillsdon:
Pending discussion.


microbit-matt-hillsdon:
It's now officially "Python Editor" when not referring to the generic idea of a Python editor.

Closing.

PR #851, opened 2022-07-28
Adjust split view on simulator to match design.

microbit-robert:
Prevent pointer interactions in iframe when dragging.


microbit-matt-hillsdon (approved):
LGTM. No change without flag and a step in the right direction for the sim.

PR #852, opened 2022-07-29
Add serial for simulator (very incomplete; behind simulator flag)

microbit-robert:
This also applies some UI changes in line with the designs.


src/device/device-hooks.tsx
```diff
@@ -43,6 +44,17 @@ export const useDevice = () => {
   return device;
 };
 
+/**
+ * Hook to access the simulator from UI code.
+ */
+export const useSimulatorDevice = (): SimulatorDeviceConnection => {
```
microbit-matt-hillsdon, line 50: We can call this useSimulator now the other hook has gone.

src/device/simulator.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  clearDevice(): void {
+    this.setStatus(ConnectionStatus.NO_AUTHORIZED_DEVICE);
+  }
+
+  private postMessage(kind: string, data: any): void {
+    const iframe = this.iframe();
+    if (!iframe) {
+      throw new Error("No-one to talk to.");
```
microbit-matt-hillsdon, line 136: Probably more whimsical than is prudent.

src/serial/SerialArea.tsx
```diff
@@ -14,6 +14,7 @@ import XTerm from "./XTerm";
 interface SerialAreaProps extends BoxProps {
   compact?: boolean;
   onSizeChange: (size: "compact" | "open") => void;
+  showTerminal?: boolean;
```
microbit-matt-hillsdon, line 17: We should decide whether we need this now. It might be better for sim serial not to come and go? But then I guess it'll need a disabled state. Worth some discussion but can be after this PR as its location is very much not final too.

src/device/simulator.ts
microbit-robert: [`03883e5`](https://github.com/microbit-foundation/python-editor-next/pull/852/commits/03883e55f0d0872a930d4e854b722a43ef75e7fc)

microbit-matt-hillsdon (approved):
Looks good. The DeviceConnection abstraction is worth a bit more thought at some point, let's discuss that when we have a stopped state in the sim.

PR #858, opened 2022-08-03
Add stop and reset to simulator.

microbit-robert:


src/simulator/SimulatorActionBar.tsx
```diff
@@ -4,38 +4,70 @@
...
-        variant="solid"
-        onClick={handlePlay}
-        icon={<RiPlayFill />}
-        aria-label="Run"
-      />
+      {syncStatus === SyncStatus.OUT_OF_SYNC ||
+      simState === SimState.STOPPED ? (
+        <IconButton
+          size={size}
+          variant="solid"
```
microbit-matt-hillsdon, line 51: Maybe we should drop this now? None of the designs have had a primary button here.
If it ends up just being stop that's an even stronger case for it not being primary.

microbit-robert: Agreed and aligned with the designs - [`bac8a8c`](https://github.com/microbit-foundation/python-editor-next/pull/858/commits/bac8a8c6b016740b7997e76b1a2c105e9a2a6d9e).

PR #859, opened 2022-08-03
Hide sync status for simulator serial.

microbit-robert:
The sync status is no longer required for the simulator serial bar as the simulator is obviously in sync when running and out of sync when stopped. The reset button is disabled when the simulator is stopped.


microbit-matt-hillsdon (approved):
LGTM

PR #861, opened 2022-08-04
Simulator padding and max width.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
Max width is somewhat arbitrary but it's a step in the right direction, I think. Intent is to cap the height to leave space for what's below.

microbit-robert (approved):
LGTM.

PR #865, opened 2022-08-05
Improve serial area in the simulator.

microbit-matt-hillsdon:
- Decrease font size
- Flip expand/collapse icon to match behaviour in sim
- Remove info button as no space (may revisit or move to menu)


microbit-robert (approved):
LGTM.

PR #866, opened 2022-08-05
Update simulator controls to make use of iframe play button.

microbit-robert:


src/simulator/SimulatorActionBar.tsx
```diff
@@ -56,27 +56,23 @@ const SimulatorActionBar = (props: SimulatorActionBarProps) => {
     }
     setIsMuted(!isMuted);
   }, [device, isMuted, setIsMuted]);
+  useEffect(() => {
+    device.on(EVENT_REQUEST_FLASH, handlePlay);
+    return () => {
+      device.removeListener(EVENT_REQUEST_FLASH, handlePlay);
+    };
+  }, [device, handlePlay]);
```
microbit-matt-hillsdon, line 64: We could consider doing this in simulator.ts but no strong view. Happy to merge as is.

microbit-matt-hillsdon (approved):
Looks good.

src/simulator/SimulatorActionBar.tsx
microbit-matt-hillsdon: Rob pointed out that we don't have access to the file system from the device. Let's leave it as is.

Issue #864, opened 2022-08-05
SoundEvent API descriptions seem reversed


martinwork:
## Bug Description

These descriptions seem reversed

![image](https://user-images.githubusercontent.com/989126/183058891-c2aef808-a6b0-4327-a913-55b83f760b7e.png)

microbit-matt-hillsdon:
Thanks, good catch. I'll fix these later today along with the spi feedback.

microbit-matt-hillsdon:
Fixed in main branch. Redeploying to the beta now (~10 mins).

I'll update the API documentation in Crowdin at ~5pm.

martinwork:
`Thanks Matt!

microbit-matt-hillsdon:
Updated in Crowdin.

PR #867, opened 2022-08-05
Update typeshed with stub fixes.

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-next/issues/864


microbit-matt-hillsdon:
Confirmed changed doc strings are correct in the deployment. Will merge after e2e completes.

PR #868, opened 2022-08-08
Add UI to show/hide simulator.

microbit-robert:


microbit-robert:
Box shadow is going to be a pain. Could move show/hide buttons into split view divider, but left hand button won't shift position with the header height change on small screens.

src/common/SplitView/HideSplitViewButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+          position="absolute"
+          left="-10px"
+          top="-5px"
+        />
+      )}
+      <CollapsibleButton
+        mode={text ? "button" : "icon"}
+        text={text}
+        icon={<Icon as={RiDownloadLine} transform={rotation} />}
+        fontSize="xl"
```
microbit-matt-hillsdon, line 55: lg to match the buttons below?

Otherwise, visually I really like this. Will look more closely at the styling tomorrow.

microbit-robert:
Currently lacks box-shadow for left hand divider due to stacking context difficulties and the need to retain hiding the collapse button when search is opened.

src/editor/EditorArea.tsx
```diff
@@ -59,7 +69,20 @@ const EditorArea = ({
           filename={selection.file}
           onSelectedFileChanged={onSelectedFileChanged}
         />
-        <ZoomControls display={["none", "none", "none", "flex"]} />
+        <Flex alignItems="center">
+          <ZoomControls display={["none", "none", "none", "flex"]} />
+          {flags.simulator && !simulatorShown && (
+            <HideSplitViewButton
+              aria-label="Show simulator"
```
microbit-matt-hillsdon, line 76: Needs translation. I guess we just add this.

microbit-matt-hillsdon: Or we try the similator.en.json approach.

src/editor/EditorArea.tsx
```diff
@@ -59,7 +69,20 @@ const EditorArea = ({
...
+              handleClick={showSimulator}
+              splitViewShown={simulatorShown}
+              direction="expandLeft"
+              text="Simulator"
```
microbit-matt-hillsdon, line 80: Likewise.

microbit-matt-hillsdon:
Last issue is that you can tab into the collapsed sim from the header:

image

PR #870, opened 2022-08-09
Add microphone.

microbit-matt-hillsdon:
In future it'll have more than just the sound level (recording) so keep
the section named differently from the sensor.


src/simulator/SimulatorModules.tsx
```diff
@@ -7,25 +7,33 @@ import { RangeSensor as RangeSensorType, Sensor } from "./model";
 import RangeSensor from "./RangeSensor";
 
 import { IconType } from "react-icons";
-import { RiSunFill, RiTempHotFill } from "react-icons/ri";
+import {
+  RiSunFill,
+  RiTempHotFill,
+  RiVolumeUpLine,
```
microbit-robert, line 13: Remove unused import.

PR #871, opened 2022-08-09
Simulator: Reset serial before flash.

### LINKED ISSUE #869: Simulator terminal error is not cleared.
To reproduce:

- Run simulator with no code. 
- Create error, e.g., use undefined variable.
- Run simulator and see error output in simulator terminal.
- Clear error in code.
- Re-run simulator and see that error output is not cleared.

microbit-matt-hillsdon:
Matches the real device and prevents old tracebacks being shown in the
serial area in the sim.

Fixes #869


microbit-robert (approved):
LGTM.

PR #873, opened 2022-08-09
Switch focus to simulator when expanded.

microbit-robert:
This now mimics the tab focus behaviour for the left hand sidebar on expand and collapse.

### DIFF
--- a/src/simulator/Simulator.tsx
+++ b/src/simulator/Simulator.tsx
@@ -16,9 +16,14 @@ import SimulatorSplitView from "./SimulatorSplitView";
 interface SimulatorProps {
   simulatorShown: boolean;
   setSimulatorShown: React.Dispatch<React.SetStateAction<boolean>>;
+  showSimulatorButtonRef: React.RefObject<HTMLButtonElement>;
 }
 
-const Simulator = ({ simulatorShown, setSimulatorShown }: SimulatorProps) => {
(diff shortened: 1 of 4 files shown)


src/simulator/Simulator.tsx
```diff
@@ -39,6 +39,15 @@ const Simulator = ({ simulatorShown, setSimulatorShown }: SimulatorProps) => {
   const hideSimulator = useCallback(() => {
     setSimulatorShown(false);
   }, [setSimulatorShown]);
+  useEffect(() => {
+    if (simulatorShown) {
+      ref.current!.focus();
+    } else {
+      (
+        document.querySelector("#simulator-expand") as HTMLButtonElement
```
microbit-matt-hillsdon, line 47: Not sure I like the action-at-a-distance nature of this. Wondering how it would be with a ref.

microbit-robert: Switch to using a ref here -> [`3e08dc7`](https://github.com/microbit-foundation/python-editor-next/pull/873/commits/3e08dc7ba7c3b89cb54aa6098236f9dbd7e02682). It adds a little noise, but it's not as bad as I thought it would be. Everything will be a lot cleaner once the simulator flag is no longer needed.

microbit-matt-hillsdon (approved):
LGTM.

PR #874, opened 2022-08-10
Improve sidebar and simulator starting widths.

microbit-robert:


src/workbench/Workbench.tsx
```diff
@@ -108,6 +127,7 @@ const Workbench = () => {
                 setSerialStateWhenOpen={setSerialStateWhenOpen}
                 setSimulatorShown={setSimulatorShown}
                 simulatorShown={simulatorShown}
+                collapseSidebar={collapseSidebar}
```
microbit-matt-hillsdon, line 130: showSimulator and showSidebar props might be easier to follow than this slightly odd pair

microbit-matt-hillsdon:
I think we should consider flipping it so the simulator is open by default on smaller screens but not Reference (if simulator flag is on).

microbit-robert:
> I think we should consider flipping it so the simulator is open by default on smaller screens but not Reference (if simulator flag is on).

After the latest commit, this should be the behaviour:

With simulator flag:

- Simulator always shown.
- Reference section hidden on smaller screen widths.

Without simulator flag:

- Reference section hidden on smaller screen widths.

src/workbench/Workbench.tsx
microbit-matt-hillsdon: I want to rejig this to avoid the sim and the sidebar components knowing about each other. We should be able to pull this behaviour up into Workbench where the layout belongs.

microbit-matt-hillsdon:
Closed in favour of #880

PR #880, opened 2022-08-11
Layout defaults with and without the sim.

microbit-matt-hillsdon:
Toggle one off and the other on as needed to preserve space.
Initial space is still tight.
You can also lose buttons as we never hide on window resize.


microbit-robert (approved):
LGTM. Starting positions and widths all look good. Collapse and expand behaviour is as expected.

Issue #877, opened 2022-08-11
Gutter width does not increase with font size.


microbit-robert:
The gutter width does not increase with font size which causes gutter markers to be truncated at font sizes of 20 or more.

![image](https://user-images.githubusercontent.com/95928279/184108066-51b065b8-fbb0-4c5b-acd9-c85e4259408f.png)

microbit-matt-hillsdon:
Lint markers have 1em width height.
Width is 1.8rem = 28.8px 
Font size is, somewhat improbably, 16pt.
So I make that 1.35em for an equivalent gutter width.

Need to find .cm-gutter.cm-gutter-lint in the theme.

PR #881, opened 2022-08-11
Lint gutter width scales with font size.

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-next/issues/877


microbit-matt-hillsdon:
Confirmed initial gutter size is identical in incognito.

microbit-robert (approved):
LGTM.

PR #882, opened 2022-08-11
Add buttons module for simulator.

microbit-robert:


microbit-robert:
Effectively a working prototype of the current button design/behaviour without the touch logo. Space is quite tight, I'm not sure we'd get all three on a single row with the current simulator width. 

We currently disable the buttons when the simulator is not running and reset any 'button toggled' state as this matches the current dispose behaviour of the `ButtonUI` in the simulator.

microbit-robert:
Issue to resolve:

- If the button is toggled as 'held down' in the Editor UI, the user can still press the button as normal on the SVG board in the iframe. This causes the iframe and Editor controls to fall out of sync.

src/simulator/ButtonModule.tsx
```diff
@@ -0,0 +1,131 @@
...
+    };
+    const mouseUpListener = (event: MouseEvent) => {
+      event.preventDefault();
+      handleSensorChange(sensor.min);
+    };
+    const mouseLeaveListener = () => {
+      handleSensorChange(sensor.min);
+    };
+    if (buttonEl) {
+      buttonEl.addEventListener("mousedown", mouseDownListener);
```
microbit-matt-hillsdon, line 89: Maybe we should rewrite these to use react's event model as there's probably no reason to drop down to HTML here.

microbit-robert: [`838a4a2`](https://github.com/microbit-foundation/python-editor-next/pull/882/commits/838a4a2e0879e724338d5b6bad1b517e323d9f03).

src/simulator/ButtonModule.tsx
```diff
@@ -50,81 +50,86 @@ const SensorButton = ({
...
-    if (disabled) {
-      setIsPressedOverride(false);
+    if (sensor.value === sensor.min) {
+      setIsHeld(false);
+    }
+    if (sensor.value === sensor.max) {
+      setIsHeld(true);
     }
-  }, [disabled]);
+  }, [sensor]);
```
microbit-matt-hillsdon, line 115: Can this be done in an event handler?

microbit-robert: I think we need the effect because we listen to sensor changes resulting from interaction with the SVG board as well as the in Editor buttons/switches?

microbit-matt-hillsdon: Ah, I see what you mean. Discussing a possible alternative on chat but fine with it like this.

microbit-robert: Contrary to what I said in chat earlier, this appears to be better to the `useEffect` alternative -> [`720cfb6`](https://github.com/microbit-foundation/python-editor-next/pull/882/commits/720cfb67826e97250951cb97ed0d6498295d7344).

microbit-matt-hillsdon: I guess the alternative is to pull the button state up so it can be reconciled with the sensor state at the point where it changes (+ some change to the model to store whether it's held). This is fine though as it keeps it local to buttons.

Thanks.

PR #883, opened 2022-08-12
Hide sidebar on resize.

microbit-matt-hillsdon:
Maintains visibility of the project action bar buttons.


microbit-robert (approved):
LGTM with and without sim.

PR #886, opened 2022-08-15
Add minimised simulator sensors

microbit-robert:


microbit-robert:
Requires icons to complete alignment with design. Merging in order to continue with adding simulator strings to `simulator.en.json`.

PR #889, opened 2022-08-16
Add pins sensor module for simulator

microbit-robert:


microbit-robert:
This PR adds temporary icons for buttons and pins to get an impression of width restrictions on the minimised form of pins.

PR #890, opened 2022-08-17
Add simulator e2e tests

microbit-robert:


microbit-robert:
Multiple files test fixes were due to not needing to confirm project replacements if the default program has not been changed.

PR #891, opened 2022-08-18
Update simulator sensor icons

microbit-robert:


microbit-robert:
If we're going to use the 'button held' icon, we'll also need an equivalent for a holding a pin for consistency.

PR #898, opened 2022-08-22
Set panel focus on link to reference from sim.

microbit-robert:


src/simulator/SimulatorModules.tsx
```diff
@@ -155,6 +155,13 @@ const CollapsibleModule = ({
   const disclosure = useDisclosure();
   const intl = useIntl();
   const [, setRouterState] = useRouterState();
+  const setPanelFocus = () => {
+    const sidebarElement = document.querySelector("[aria-label='Sidebar']");
```
microbit-matt-hillsdon, line 159: Querying globally like this is fragile and presumably breaks if you change language.

I think this should be part of the navigation state.

src/workbench/SideBar.tsx
```diff
@@ -124,17 +124,26 @@ const SideBar = ({
     if (tabIndex !== -1) {
       onTabIndexChange(tabIndex);
       onSidebarExpand();
-      if (!api && !reference && !idea) {
+      if ((!api && !reference && !idea) || focusTab) {
```
microbit-robert, line 127: I don't remember why we need this -> `!api && !reference && !idea`? If we didn't do this check, we wouldn't need `focusTab`.

src/simulator/SimulatorModules.tsx
microbit-robert: Added `focusTab` to navigation state here -> [`f993ac8`](https://github.com/microbit-foundation/python-editor-next/pull/898/commits/f993ac8d7f1e320517d1ad4a94e757f16f97dfff).

src/simulator/SimulatorModules.tsx
```diff
@@ -181,9 +181,11 @@ const CollapsibleModule = ({
       {
         tab: "reference",
         reference: { id: references[id] },
+        focusTab: true,
       },
       "documentation-from-simulator"
     );
+    // setPanelFocus();
```
microbit-matt-hillsdon, line 188: ?

src/workbench/SideBar.tsx
microbit-matt-hillsdon: My initial thought is that we were attempting to leave focus management to the code that handles the anchor for a specific item. But... that code doesn't actually focus. So I'm tentatively on board with removing the guard. I'll try it out.

microbit-matt-hillsdon: Removing these breaks focusing the panel when clicking a tab.

microbit-matt-hillsdon: I've kept the focus as part of the navigation state. I've tried to switch to using the identity of the params as an indication of when a navigation has occurred so we can know it's safe to focus if requested.

microbit-matt-hillsdon:
Bug:
- Open API
- Pick section
- Attempt to collapse sidebar (fails)
- Attempt to collapse sidebar (works)

microbit-matt-hillsdon:
I gave up on trying to improve this. I really think it could be better but it's too risky for this point.

In particular it would help if tabIndex was derived from tab and sidebar visibility and perhaps if the sidebar/simulator visibility was managed by the router state. Too large a change for this point.

Minor tweak to avoid serializing "focus" in the URL.

PR #900, opened 2022-08-22
Prevent the show simulator button from stealing focus in some situations.

### LINKED ISSUE #893: Show simulator button focused when Reference opened
On small screens, if you open the Reference sidebar while the simulator is open, the simulator is closed, but the focus is moved to the 'show simulator' button instead of inside the Reference tabpanel.

microbit-robert:
Closes #893.


microbit-robert:
I don't like introducing the extra state, but can't think of an alternative way to do it without having to wait for the 'show simulator' button to be available on the page.

microbit-matt-hillsdon:
> I don't like introducing the extra state, but can't think of an alternative way to do it without having to wait for the 'show simulator' button to be available on the page.

Took me a while to understand but I think this makes sense. I think ideally they'd be the same button in the show and hide case (not tried this!) and then no special focus management would be necessary.

PR #902, opened 2022-08-23
Simplify the range sensor thresholds.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
Good to merge assuming CI passes.

PR #907, opened 2022-08-23
Simulator serial tweaks.

microbit-matt-hillsdon:
- Show the hints and tips entry for sim serial, but on the menu due to
  lack of space.
- Show the expand/collapse text, but deconflict with the traceback on
  the sim due to lack of space. We decided this is important to
  understand that it is the serial area as it's otherwise unlabelled.


microbit-matt-hillsdon:
Change looks good.

PR #909, opened 2022-08-25
Adapt to simulator API change

microbit-matt-hillsdon:
The state now contains some entries that aren't sensors and has changed shape to an object with a property per entry.
Top-level state entries are always updated as a whole.
The messages have changed to reflect this model.


src/device/simulator.ts
```diff
@@ -114,9 +190,25 @@ export class SimulatorDeviceConnection
     });
   }
 
-  sensorWrite = async (id: string, value: number): Promise<void> => {
-    this.postMessage("sensor_set", {
-      sensor: id,
+  setSimulatorValue = async (
+    id: string,
```
microbit-matt-hillsdon, line 194: This can probably be SensorStateKey now. Nothing tracks the sensor types but not sure that's worth it.

src/simulator/AccelerometerModule.tsx
```diff
@@ -2,42 +2,61 @@ import { HStack, IconButton, Select, Stack, Text } from "@chakra-ui/react";
...
+  SimulatorState,
+} from "../device/simulator";
 import RangeSensor from "./RangeSensor";
 
 interface AccelerometerModuleProps {
   icon: ReactNode;
-  sensors: Record<string, Sensor>;
-  onSensorChange: (id: string, value: any) => void;
+  state: SimulatorState;
+  onValueChange: (id: string, value: any) => void;
```
microbit-matt-hillsdon, line 15: Simularly, the type of onValueChange's id parameter can change.

src/simulator/Simulator.tsx
```diff
@@ -90,7 +90,7 @@ const Simulator = ({
                 ref={ref}
                 as="iframe"
                 // This needs changing before we remove the flag.
-                src={`https://stage-python-simulator.microbit.org/simulator.html?color=${encodeURIComponent(
+                src={`https://review-python-simulator.microbit.org/api-change/simulator.html?color=${encodeURIComponent(
```
microbit-matt-hillsdon, line 93:
```suggestion
                src={`https://stage-python-simulator.microbit.org/simulator.html?color=${encodeURIComponent(
```

microbit-matt-hillsdon: This needs to revert back before merge. Leaving for a last check.

microbit-matt-hillsdon: All looks good. Reverting this (which will break the preview).

PR #924, opened 2022-09-01
[Simulator] Add missed compass module UI strings.

microbit-robert:


microbit-matt-hillsdon:
Code change looks good. As discussed Rob will do a a final sweep for sim strings today.
Will wait until it builds to check link then approve.

microbit-matt-hillsdon (approved):

PR #927, opened 2022-09-01
Flip mute/unmute icons.

microbit-matt-hillsdon:
You can find examples of this either way, but this matches MakeCode and
we've already had feedback about this.

Change the icon to look louder.


microbit-robert (approved):
LGTM.

PR #931, opened 2022-09-03
Early preview of French translations

microbit-matt-hillsdon:


microbit-matt-hillsdon:
Superseded by #933. 

The deployment was shared with a translator but it will continue to work after the PR is closed.
We hope to merge #933 before too long.

PR #934, opened 2022-09-05
Fix ICU syntax.

microbit-matt-hillsdon:


microbit-matt-hillsdon:
I'll update the UI strings in Crowdin once this is merged. It should be the only change.

microbit-robert (approved):
LGTM.

PR #936, opened 2022-09-07
Tweak English names of languages.

microbit-matt-hillsdon:
Match Google Translate's presentation.

### DIFF
--- a/src/settings/LanguageDialog.tsx
+++ b/src/settings/LanguageDialog.tsx
@@ -70,6 +70,10 @@ export const LanguageDialog = ({ isOpen, onClose }: LanguageDialogProps) => {
                 <Icon as={RiExternalLinkLine} />
               </Link>
             </VStack>
+            <Text fontSize="xs" alignSelf="flex-end" mt={1}>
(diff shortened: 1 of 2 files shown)


src/settings/LanguageDialog.tsx
```diff
@@ -70,6 +70,10 @@ export const LanguageDialog = ({ isOpen, onClose }: LanguageDialogProps) => {
                 <Icon as={RiExternalLinkLine} />
               </Link>
             </VStack>
+            <Text fontSize="xs" alignSelf="flex-end" mt={1}>
+              * These languages are an early preview of in-progress
```
microbit-robert, line 74: Maybe this should now have the "intentionally not translated" comment.

microbit-robert (approved):
Looks good. One very minor comment.

src/settings/LanguageDialog.tsx
microbit-matt-hillsdon: Makes sense, though I'll revisit and/or remove this before launch so not going to bother now.

PR #939, opened 2022-09-07
Fix for simulator radio output bug

microbit-robert:


microbit-matt-hillsdon:
Thanks, good catch.

microbit-matt-hillsdon (approved):
LGTM.

PR #942, opened 2022-09-08
Accessibility improvements

### LINKED ISSUE #941: Can't tab out of the serial terminal
There doesn't appear to be a way to tab out of the serial terminal forwards. This is more problematic in the simulator area where the user will want to tab forward to access the simulator sensors and controls.

Requires investigation, but we may need to implement a custom shortcut using `Esc + Tab` to exit the terminal. This would keep the shortcut consistent with CodeMirror.

microbit-robert:
- Adds the ability to tab out of the serial terminals.
- Retains focus on the gesture send button after the timeout.

Closes #941.

### DIFF
--- a/src/serial/xterm-keyboard.ts
+++ b/src/serial/xterm-keyboard.ts
@@ -12,6 +12,24 @@ const copyShortcut = (e: KeyboardEvent): boolean => {
   return true;
 };
 
+let escWasPressed = false;
+let escTimeout: NodeJS.Timeout;
+const tabOutShortcut = (e: KeyboardEvent, tabTo: HTMLElement): boolean => {
+  if (e.code === "Escape") {
+    clearTimeout(escTimeout);
+    escWasPressed = true;
+    escTimeout = setTimeout(() => {
+      escWasPressed = false;
+    }, 500);
+  }
+  if (e.code === "Tab" && escWasPressed) {
+    e.preventDefault();
+    tabTo.focus();
+    return false;
+  }
+  return true;
+};
+
 const isMicroPythonCtrlShortcut = (code: string) => {
   // Editing keys are here: https://github.com/micropython/micropython/blob/46a11028521425e7d6c85458c849bb96ff82152e/lib/mp-readline/readline.c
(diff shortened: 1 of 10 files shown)


src/serial/xterm-keyboard.ts
```diff
@@ -12,6 +12,24 @@ const copyShortcut = (e: KeyboardEvent): boolean => {
   return true;
 };
 
+let escWasPressed = false;
+let escTimeout: NodeJS.Timeout;
```
microbit-matt-hillsdon, line 16: I think just type this as any - I've found these typings to be a real pain over time as sometimes you find NodeJS.Timeout is allowed and sometimes not depending on what you have installed.

src/serial/xterm-keyboard.ts
```diff
@@ -12,6 +12,24 @@ const copyShortcut = (e: KeyboardEvent): boolean => {
...
+const tabOutShortcut = (e: KeyboardEvent, tabTo: HTMLElement): boolean => {
+  if (e.code === "Escape") {
+    clearTimeout(escTimeout);
```
microbit-matt-hillsdon, line 19: Shouldn't we also clear the timeout for non-escape key presses (including tab)?

microbit-robert: I've tweaked this to clear the timeout and reset `escWasPressed` to `false` if any key other than `escape` was pressed. Pressing `escape` should just clear and reset the timeout. This means pressing `escape`, `*any other key*` , then `tab` will not exit the terminal. [`9d5c827`](https://github.com/microbit-foundation/python-editor-next/pull/942/commits/9d5c82796d502173b7ceb681a8db7e5b70e1f2d6)

microbit-matt-hillsdon: Thanks, this looks good.

microbit-matt-hillsdon: I want to have a quick play and convince myself the React context stuff is necessary, otherwise this looks good.

PR #943, opened 2022-09-08
Add search support for multiple languages

microbit-robert:
The search function is currently not supported for Korean.

### DIFF
--- a/src/documentation/search/search.ts
+++ b/src/documentation/search/search.ts
@@ -4,7 +4,9 @@
  * SPDX-License-Identifier: MIT
  */
 import lunr from "lunr";
+import multi from "lunr-languages/lunr.multi";
 import stemmerSupport from "lunr-languages/lunr.stemmer.support";
+import tinyseg from "lunr-languages/tinyseg";
 import { retryAsyncLoad } from "../../common/chunk-util";
 import { splitDocString } from "../../editor/codemirror/language-server/docstrings";
 import type {
@@ -23,6 +25,9 @@ import {
 import { contextExtracts, fullStringExtracts, Position } from "./extracts";
 
 stemmerSupport(lunr);
+multi(lunr);
+// Required for Ja stemming support.
+tinyseg(lunr);
 
 const ignoredPythonStopWords = new Set([
   // Sorted.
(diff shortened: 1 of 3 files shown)


src/documentation/search/search.ts
```diff
@@ -200,12 +205,20 @@ const apiSearchableContent = (
...
   tab: "reference" | "api",
+  language: string,
   ...plugins: lunr.Builder.Plugin[]
 ): SearchIndex => {
   const index = lunr(function () {
     this.ref("id");
     this.field("title", { boost: 10 });
     this.field("content");
+    // There is always some degree of English content.
+    const multiLanguages = ["en"];
```
microbit-robert, line 216: I've seen some GitHub issues where using multiple languages causes issues with tokenisation of the main target language. It might be more sensible to throwaway English here unless the main language is not supported. https://github.com/MihaiValentin/lunr-languages/issues

microbit-robert:
There is PR for Korean support -> https://github.com/MihaiValentin/lunr-languages/pull/81

src/documentation/search/search.ts
microbit-matt-hillsdon: Can you try to get to the bottom of this? Is it working for Japanese? I read https://github.com/MihaiValentin/lunr-languages/issues/45 -  are there others that are relevant?

microbit-robert: I also only found the issue for Japanese. 

After running tests, it looks like we find "日本語" in the following  "この文章は日本語で書かれています" in both the single ja pipeline and the multi-language en and ja pipeline. So the multilang approach is not a problem.

The issue we retain is that searching for "書かれ" returns nothing, but "書か れ" is a match. Indexing splits "書かれ" into "書か" and "れ". I'm not sure how we deal with this.

microbit-matt-hillsdon: OK, we'll merge this as a big step forward and I'll try to look further at this.

src/documentation/search/search.ts
```diff
@@ -252,6 +266,31 @@ async function loadLunrLanguageSupport(
...
+    case "ja":
+      return (await import("lunr-languages/lunr.ja")).default;
+    default:
+      // No search support for the language, default to lunr's built-in English support.
+      return undefined;
+  }
+}
+
+function convertLangToLunrParam(language: string): string | undefined {
+  // Korean is not supported by lunr-languages.
```
microbit-matt-hillsdon, line 283: We'll also look at this in a subsequent PR.

PR #944, opened 2022-09-09
Update translations.

microbit-matt-hillsdon:
Includes changes from:
https://github.com/microbit-foundation/micropython-microbit-stubs/pull/65
https://github.com/microbit-foundation/pyright/pull/20


microbit-robert (approved):
Looks good.  zh-tw changes have reverted translated API names back to English. Not important as we will hide them, but worth noting.

Issue #947, opened 2022-09-09
Simulator accessibility issues


microbit-robert:
Screen readers:

- The sound level thresholds are read as "clickable".
- It is not clear if the reader is reading messages sent from the micro:bit, or messages sent to the micro:bit by the user.
- The compass heading slider in the expanded compass module is missing a label. The reader reads "slider 0", rather than "compass heading slider 0".
- The reader reads the compass heading slider value in the collapsed compass module twice. E.g., "compass heading slider 00". Edit: This is because the tooltip is being read as well as the input value separately.

microbit-robert:
Closed by #948.

PR #948, opened 2022-09-09
Simulator accessibility improvements.

microbit-robert:
See #947.


src/simulator/RangeSensor.tsx
```diff
@@ -161,6 +162,7 @@ const ThresholdMark = ({ value, label, min, max }: ThresholdMarkProps) => {
   return (
     <Tooltip hasArrow placement="top" label={intl.formatMessage({ id: label })}>
       <Box
+        aria-label={intl.formatMessage({ id: label }) + ` ${value}`}
```
microbit-robert, line 165: This feels horrible, but would need changes to the UI strings to do properly?

src/simulator/RangeSensor.tsx
```diff
@@ -158,9 +163,11 @@ interface ThresholdMarkProps {
 const ThresholdMark = ({ value, label, min, max }: ThresholdMarkProps) => {
   const intl = useIntl();
   const percentLeft = ((value - min) / (max - min)) * 100 + "%";
+  const formattedLabel = intl.formatMessage({ id: label }) + ` ${value}`;
```
microbit-robert, line 166: Is this sensible, or does it need a new UI string to be formatted correctly?

lang/ui.en.json
```diff
@@ -903,6 +903,10 @@
     "defaultMessage": "Radio",
     "description": "Radio simulator panel title"
   },
+  "simulator-radio-code": {
+    "defaultMessage": "Radio message from the micro:bit simulator",
```
microbit-matt-hillsdon, line 907: These feel like they might be frustratingly long. One to discuss.

src/simulator/RangeSensor.tsx
```diff
@@ -83,8 +85,11 @@ const RangeSensor = ({
           placement="top"
           label={valueText}
           isOpen={minimised ? showTooltip : false}
+          onOpen={() => sliderRef.current!.removeAttribute("aria-describedby")}
```
microbit-matt-hillsdon, line 88: This feels risky, presumably this is react-managed DOM.

microbit-matt-hillsdon: Is there any value we can pass in for the prop that will achieve this?

lang/ui.en.json
microbit-robert: Agreed.

src/simulator/RangeSensor.tsx
microbit-matt-hillsdon: Different approach in 40b13963703dc6d07231a0b1483d9aaa6b3d019c. Instead of removing it at runtime (which might break reconciliation), wrap the component and ignore it.

microbit-matt-hillsdon (approved):

PR #955, opened 2022-09-13
Move the beta notice to a flag.

microbit-matt-hillsdon:
Enable it for non-production stages.

This removes livePreview. To get the same effect you need to do flag=none.

We can merge this now, no need to wait for the release.

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -8,4 +8,3 @@ export const version = process.env.REACT_APP_VERSION || "local";
 export type Stage = "local" | "REVIEW" | "STAGING" | "PRODUCTION";
 
 export const stage = (process.env.REACT_APP_STAGE || "local") as Stage;
-console.log("Stage", stage);


src/environment.ts
```diff
@@ -4,4 +4,8 @@
  * SPDX-License-Identifier: MIT
  */
 export const version = process.env.REACT_APP_VERSION || "local";
-export const stage = process.env.REACT_APP_STAGE || "local";
+
+export type Stage = "local" | "REVIEW" | "STAGING" | "PRODUCTION";
```
microbit-matt-hillsdon, line 8: Typed this to reduce chance of STAGE vs STAGING screw ups in the flag config.

microbit-matt-hillsdon:
Tempted to remove the ability to enable all flags. As we use them more for environment differences like this it'll cease to make sense.

src/environment.ts
```diff
@@ -4,4 +4,8 @@
...
+
+export const stage = (process.env.REACT_APP_STAGE || "local") as Stage;
+console.log("Stage", stage);
```
microbit-robert, line 11: Remove debug.

microbit-robert:
> Tempted to remove the ability to enable all flags. As we use them more for environment differences like this it'll cease to make sense.

Yep, that makes sense. I don't remember the last time I used `flag=*`.

microbit-robert (approved):
One minor comment, otherwise this looks good. Also in favour or removing the all flags flag.

src/environment.ts
microbit-matt-hillsdon, line 11:
```suggestion
```

microbit-matt-hillsdon:
> Also in favour or removing the all flags flag.

I'll do this separately at some point.

Issue #956, opened 2022-09-13
Korean translation issue with delete-file-action


microbit-matt-hillsdon:
![image](https://user-images.githubusercontent.com/44397098/189927814-6ae219bd-9d08-4fc4-bb88-2d473bfcd006.png)

When running locally you get a useful logged error that we don't get in review deployments. I'll check all other languages locally too.

```
utils.js:19 Error: [@formatjs/intl Error FORMAT_ERROR] Error formatting message: "delete-file-action", using id as fallback. 
Locale: ko
MessageID: delete-file-action
Default Message: undefined
Description: undefined 
 

The intl string context variable "filename" was not provided to the string "undefined"
Error: The intl string context variable "filename" was not provided to the string "undefined"
    at MissingValueError.FormatError [as constructor] (error.js:14:1)
    at new MissingValueError (error.js:44:1)
    at formatToParts (formatters.js:66:1)
    at IntlMessageFormat.formatToParts (core.js:117:1)
    at IntlMessageFormat.format (core.js:95:1)
    at formatMessage (message.js:79:1)
    at formatMessage (provider.js:44:1)
    at FormattedMessage (message.js:21:1)
    at renderWithHooks (react-dom.development.js:14985:1)
    at mountIndeterminateComponent (react-dom.development.js:17811:1)
    at new MessageFormatError (error.js:50:1)
    at formatMessage (message.js:82:1)
    at formatMessage (provider.js:44:1)
    at FormattedMessage (message.js:21:1)
```

microbit-matt-hillsdon:
> Same string has an issue in zh-CN.

No, it doesn't, this is just a very confusing message from formatjs when switching from ko to zh-CN.

microbit-matt-hillsdon:
Fixed the string in Crowdin: https://crowdin.com/translate/microbitorg/6426/en-ko#331470

I'll re-do the translation sync and discuss further string QA elsewhere.

PR #958, opened 2022-09-13
Check UI string variables match.

microbit-matt-hillsdon:
Would have caught https://github.com/microbit-foundation/python-editor-next/issues/956 No additional errors.
I've also manually reviewed the ICU plural strings, and while a surprising number don't seem to be translated there are no structural errors.

### DIFF
--- a/bin/tidy-lang.js
+++ b/bin/tidy-lang.js
@@ -8,9 +8,41 @@ const path = require("path");
 const en = JSON.parse(fs.readFileSync("lang/ui.en.json"));
 const validKeys = new Set(Object.keys(en));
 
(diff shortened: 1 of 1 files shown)


bin/tidy-lang.js
```diff
@@ -8,9 +8,41 @@ const path = require("path");
...
+        console.error(file);
+        anyError = true;
+      }
+      console.error(`  ${en}`);
+      console.error(`  ${translated}`);
+      console.error(`  Differing variables!`);
+      console.error();
+    }
+  }
+  return anyError;
```
microbit-matt-hillsdon, line 39: I've got the exit logic inverted.

PR #954, opened 2022-09-13
Update translations (preview quality).

microbit-matt-hillsdon:
Includes changes from:
- https://github.com/microbit-foundation/micropython-microbit-stubs/pull/68 (merged)
- https://github.com/microbit-foundation/pyright/pull/22 (merged)

Amount of change is significantly lower this time around.

PR #960, opened 2022-09-14
Reinstate en typeshed.

microbit-matt-hillsdon:
Unintentionally changed in
https://github.com/microbit-foundation/python-editor-next/pull/954


microbit-robert (approved):
LGTM.

microbit-matt-hillsdon:
I'll make changes to the translation update script to prevent recurrence. The issue is that we temporarily intentionally have the English content newer than the translations (from a branch) so when updating the translated typeshed it's easy to regress this. This situation should resolve itself tomorrow when commercial translation work is expected to complete and we can make a final round of changes.

PR #961, opened 2022-09-14
Prevent regressing the en stubs.

microbit-matt-hillsdon:
This is a full translation update with no change (including no change to the stubs).


microbit-robert (approved):
LGTM.

PR #963, opened 2022-09-15
[dnd] Add ability to append to code block and handle blank line spaces

### LINKED ISSUE #539: Drag and drop indentation
Without a new line with no indentation at the bottom of the file, drag and drop won't allow you to drop code beyond the last line without any indentation - it is always at the indentation level of the last line. See video below:

https://user-images.githubusercontent.com/95928279/152361196-ce4b02db-8977-4f94-b24f-5bcea0dad45e.mp4

### LINKED ISSUE #800: Drag and drop indent matching can be confused by empty blank lines
It's common for blank lines not to be indented along with the rest of the code. In any case you can't tell the difference easily. We don't special case blank lines so we can end up matching their indent level when that isn't appropriate.

E.g.

```python
while True:

    display.scroll("Drag above me note no spaces on previous line")
```

### LINKED ISSUE #812: Can't drag to append to a block
You can't drag code to append to a block unless there's a blank line with the right amount of indent at the end of the block (unlikely).

It's ambiguous whether you are closing the block or continuing it but we bias towards ending it.

The Oxford group looked at the horizontal position to determine where to snap to and this seemed to work well. We should do likewise.

### LINKED ISSUE #856: Matching indent to blank lines can be confusing
It's quite easy to end up with a whitespace indent on a blank line from a delete like this:

Screenshot2022_08_01_151321

(arguably the location of our indent lines makes this misleading)

Then if you drag content to that line we match the three spaces on the blank line not the four spaces on the line above.

That results in a pile of confusing errors:

image

Probably originally seen by Lorna but hard to be sure this repro matches.

microbit-robert:
Closes #856.
Closes #812.
Closes #800.
Closes #539.

### DIFF
--- a/src/editor/codemirror/edits.ts
+++ b/src/editor/codemirror/edits.ts
@@ -112,6 +112,21 @@ export const calculateChanges = (
         } else {
           mainFrom = state.doc.line(line).from;
         }
+        // Determine if code is being appended to previous block and indent accordingly.
+        const previousLineNum = line - 1;
+        if (previousLineNum >= 1 && previousLineNum <= state.doc.lines) {
+          const previousLine = state.doc.line(previousLineNum);
+          const currentLineIsBlank = lineIsBlank(state, line);
+          const previousLineIsBlank = lineIsBlank(state, previousLineNum);
+          mainIndent =
+            !previousLineIsBlank && currentLineIsBlank
+              ? previousLine.text.match(/^(\s*)/)?.[0] ?? ""
+              : "";
+          // Special case where body code does not yet exist but should be indented.
+          if (previousLine.text.trim().endsWith(":") && currentLineIsBlank) {
+            mainIndent += "    ";
+          }
+        }
       }
     } else {
       // When no line is specified, insert before the code (not just after the imports).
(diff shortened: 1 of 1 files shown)


microbit-robert:
Slight issue with "while smooshing". `while` is lost if appending one line below existing while block, but returns two lines below existing while block and beyond.

microbit-robert:
> Slight issue with "while smooshing". `while` is lost if appending one line below existing while block, but returns two lines below existing while block and beyond.

There appears to be a bug in the `isInWhileTrueTree` function where the tree believes it is in a `while statement` when it is not. Most easily observed when add new lines below a `while True` loop and then dragging and new `while True` example from bottom to top.

microbit-matt-hillsdon:
Rebased so it deploys properly.

microbit-matt-hillsdon:
Superseded by https://github.com/microbit-foundation/python-editor-v3/pull/1037 which avoids [this issue](https://github.com/microbit-foundation/python-editor-v3/issues/856#issuecomment-1282265325) and allow the user control over the ambiguity in https://github.com/microbit-foundation/python-editor-v3/issues/812.

PR #965, opened 2022-09-15
Update main serial font size based on zoom

microbit-robert:


microbit-matt-hillsdon (approved):
LGTM after tweaks, Rob has reviewed my changes.

PR #968, opened 2022-09-16
Attempt to reinstate a relocatable install.

microbit-matt-hillsdon:
The previous issue was worked around on https://github.com/microbit-foundation/python-editor-v3/pull/479

That fix has the consequence that we can't serve the same build from multiple URLs, which limits our deployment options.

Switch to "auto" option via WebPack (can't be set via CRA).


microbit-matt-hillsdon:
e2e tests are an easy fix.

However, this whole approach doesn't seem to be compatible with client-side routing. As we don't template index.html, we need absolute paths in the HTML file so it can be loaded from non-root locations and still find resources. I can't see an easy option that retains a relocatable app and client side routing.

microbit-matt-hillsdon:
I don't see a way around this without a small server-side piece to handle rewrites of index.html for new locations. It could be very simple, setting an HTML base tag, but not worth any risk at this point.

Closing this PR.

PR #970, opened 2022-09-19
Rationalise URLs

microbit-matt-hillsdon:
The previous scheme assumed that we'd maintain independent state for the tabs, but we decided against that.
Therefore we can go for a more traditional URL routing scheme.
It could do with some simplification but, if we can fix the animation issue then exposing these URLs would be a step forward.

### DIFF
--- a/src/router-hooks.tsx
+++ b/src/router-hooks.tsx
@@ -68,14 +68,34 @@ type RouterContextValue = [
 
 const RouterContext = createContext<RouterContextValue | undefined>(undefined);
 
-const parse = (search: string): RouterState => {
+const parse = (pathname: string, search: string): RouterState => {
   const params = new URLSearchParams(search);
-  return {
-    tab: params.get("tab") ?? undefined,
-    api: anchorForParam(params.get("api")),
-    reference: anchorForParam(params.get("reference")),
-    idea: anchorForParam(params.get("idea")),
-  };
+  if (params.get("tab")) {
+    // Legacy
+    return {
+      tab: params.get("tab") ?? undefined,
+      api: anchorForParam(params.get("api")),
+      reference: anchorForParam(params.get("reference")),
+      idea: anchorForParam(params.get("idea")),
+    };
+  }
+  const base = process.env.PUBLIC_URL ?? "/";
+  pathname = pathname.slice(base.length);
+  if (pathname) {
+    const parts = pathname.split("/");
+    const tab = parts[0];
+    switch (tab) {
+      case "api":
+        return { tab: "api", api: anchorForParam(parts[1]) };
+      case "reference":
+        return { tab: "reference", api: anchorForParam(parts[1]) };
+      case "idea":
+        return { tab: "ideas", api: anchorForParam(parts[1]) };
+      default:
+        return {};
+    }
+  }
+  return {};
 };
 
 /**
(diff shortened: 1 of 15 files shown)


src/router-hooks.tsx
```diff
@@ -68,14 +68,34 @@ type RouterContextValue = [
...
-  };
+  if (params.get("tab")) {
+    // Legacy
+    return {
+      tab: params.get("tab") ?? undefined,
+      api: anchorForParam(params.get("api")),
+      reference: anchorForParam(params.get("reference")),
+      idea: anchorForParam(params.get("idea")),
+    };
+  }
```
microbit-matt-hillsdon, line 81: We can delete this if we don't use it ourselves.

microbit-matt-hillsdon: No links here: https://github.com/microbit-foundation/microbit-org-backups/search?q=%22https%3A%2F%2Fpython.microbit.org%2Fv%2F2%22

src/router-hooks.tsx
```diff
@@ -68,14 +69,33 @@ type RouterContextValue = [
...
```
microbit-matt-hillsdon, line 82:
```suggestion
```

microbit-matt-hillsdon: Remove legacy code path as it's unused.

src/router-hooks.tsx
```diff
@@ -68,14 +69,24 @@ type RouterContextValue = [
 
 const RouterContext = createContext<RouterContextValue | undefined>(undefined);
 
-const parse = (search: string): RouterState => {
+const parse = (pathname: string, search: string): RouterState => {
   const params = new URLSearchParams(search);
```
microbit-matt-hillsdon, line 73: Remove unused:

```suggestion
```

src/router-hooks.tsx
```diff
@@ -95,28 +105,28 @@ export const useRouterState = (): RouterContextValue => {
...
-      return `${encodeURIComponent(k)}=${encodeURIComponent(
-        serializeValue(v)
-      )}`;
-    })
-    .join("&");
-  return window.location.toString().split("?")[0] + (query ? "?" + query : "");
+  // This could be cleaned up if we always set the tab.
+  const parts = [
+    state.tab ??
+      (state.tab === "api" || state.api ? "api" : undefined) ??
```
microbit-matt-hillsdon, line 111:
```suggestion
      (state.api ? "api" : undefined) ??
```

We already addressed the tab-is-set case.

microbit-matt-hillsdon:
Discussed having /reference/display/scroll and similarly for API.

Ideally we'd decorate the data so that each reference topic and API entry knew its full slug and then keep the navigation logic largely the same.

microbit-matt-hillsdon:
> Discussed having /reference/display/scroll and similarly for API.
> 
> Ideally we'd decorate the data so that each reference topic and API entry knew its full slug and then keep the navigation logic largely the same.

I gave this some thought. This is a lot of change for a relatively minor benefit. I think we'd change teh CMS model if we considered this goal up front. The slugs as used are unique, just not qualified by module (for Reference) and I think they'll be OK in practice.

microbit-robert (approved):
LGTM.

PR #971, opened 2022-09-19
Tweak the animation.

microbit-matt-hillsdon:
This seems to fix https://github.com/microbit-foundation/python-editor-v3/issues/795

I don't think it introduces any stacking context issues but please check with care.


microbit-robert (approved):
LGTM.

PR #979, opened 2022-09-21
Remove the language preview markers.

microbit-matt-hillsdon:
Keep the mechanism as we might enable it for future beta/staging releases when testing new languages.


microbit-matt-hillsdon:
@microbit-robert I propose this is the last PR we merge before we change the staging deployment location. Let me know if there's anything else we should consider as it could be late today before we can merge anything else that folks will see.

microbit-robert (approved):
LGTM.

Issue #977, opened 2022-09-21
Review tab order for accessibility


kmcnaught:
Let's review this together on a call, but I'll highlight some 

## Environment

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

kmcnaught:
Overall, when all the sidebars are open, the tab order is fairly sensible, but I think there are some improvements to be made. Some areas for discussion...

## Navigation from a closed docs sidebar
If you're loading to a small size, the left hand docs sidebar is closed, and the Ref/Ideas/API/Project are skipped from the tab order. This means you can't tab into these areas without expanding the sidebar first - and the screen reader description doesn't signpost what the contents are. 
![image](https://user-images.githubusercontent.com/12151404/191468070-551c2049-1a8a-4088-b251-487bfeea8034.png)

If instead you collapse the sidebar manually, then you get a much nicer tab order, which I think would be reasonable to copy.  
![image](https://user-images.githubusercontent.com/12151404/191467687-0649480b-dae2-42c1-a221-09f7ddea6f9e.png)

## Navigation to editor from an open docs sidebar
If the left hand sidebar is open (reference pane open by default), it takes >40 tabs to get to the editor, as you are taken via all the topic buttons. If you are using a screen reader and know all the shortcuts, you can shortcircuit by navigating by region, but it would be preferable to have a more streamlined / nested navigation order
![image](https://user-images.githubusercontent.com/12151404/191466384-fd0830fa-20ff-4db1-a3d9-254a4b4062aa.png)

## Navigation from revisited webpage
See discussion in 973 - Reading around this, NVDA by design remembers where the user's last position when returning to a previously visited web page. This isn't working quite as expected with the Python Editor.
_Originally posted by @microbit-robert in https://github.com/microbit-foundation/python-editor-v3/issues/973#issuecomment-1253422274_

See also 973

microbit-robert:
#982 has fixed the "Navigation from a closed docs sidebar" part of this issue. When using the beta deployment (inside an iFrame), NVDA is causing the focus to switch to different areas of the Editor - this does not occur when running the project locally (not in an iFrame).

Issue #973, opened 2022-09-20
Inconsistent tab order / initial focus


kmcnaught:
## Bug Description

I have been reviewing the tab order for keyboard/screen-reader users, and it appears that sometimes focus starts in the top left and follows a logical order top-to-bottom and left-to-right and other times it jumps straight into the simulator pane. Once underway the tab order is consistent, it's just like we've short-circuited it by focussing the simulator at an inopportune moment. 

## How To Reproduce

Sometimes I can reproduce this consistently by using a screen reader, or not. But I imagine this is purely coincidentally, since the page should not know whether a screen reader is being used. And it doesn't always occur consistently.

1. Load https://python.microbit.org/v/beta
2. Hit F5 to refresh to get consistent starting state
3. When page is loaded, press tab a few times and observe focussed elements

On several occasions I have seen the correct behaviour occurring when either Narrator or NVDA are running, and the incorrect behaviour occurring when they are not. As I started writing this issue I was observing the bug on the live editor, but right now I can only trigger it on a local version (2fd2bfd). 

## Expected behaviour
Tab order should start from top left like in screenshot below

### Correct tab order
Note that there may be a separate discussion to be had about the best tab order, depending on whether the left hand pane is collapsed or expanded, but this certainly seems to be the intended behaviour.

![screenshot showing tab order starting at top left of web page](https://user-images.githubusercontent.com/12151404/191240139-c7d612b9-5a75-435f-8876-b6da7ead0c7f.PNG)

### Incorrect tab order

![screenshot showing tab order starting on right-hand simulator pane](https://user-images.githubusercontent.com/12151404/191236427-a3382b9d-ef18-4ca5-a651-d7ee0813cb75.PNG)

## Environment

- OS: Windows 10
- Browser: Chrome (105.0.5195.127) and Firefox (104.0.2) 
- Version: Editor 0.1.0-dev.3084
MicroPython (micro:bit V1) 1.1.0-beta.1
MicroPython (micro:bit V2) 2.1.0-beta.3

microbit-matt-hillsdon:
It is likely that this issue was fixed by https://github.com/microbit-foundation/python-editor-v3/pull/900 merged as ac0a9ec963090f3e236ea50e2b88651fd2502c84 on September 8th vs your local commit from end of August.

kmcnaught:
[...]

PR #982, opened 2022-09-21
Fix focus bugs that occur on initial load.

microbit-robert:
This fixes some issues raised in #973 and #977.

The simulator and sidebar will no longer steal focus on load.

The sidebar tabs are always accessible, even if the sidebar loads in a collapsed state (screen widths <1366px).

### DIFF
--- a/src/simulator/Simulator.tsx
+++ b/src/simulator/Simulator.tsx
@@ -3,7 +3,14 @@
  *
  * SPDX-License-Identifier: MIT
  */
-import { AspectRatio, Box, Flex, useToken, VStack } from "@chakra-ui/react";
+import {
+  AspectRatio,
+  Box,
+  Flex,
+  usePrevious,
+  useToken,
+  VStack,
+} from "@chakra-ui/react";
 import { useEffect, useRef, useState } from "react";
 import { IntlShape, useIntl } from "react-intl";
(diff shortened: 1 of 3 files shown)


src/simulator/Simulator.tsx
```diff
@@ -67,16 +74,20 @@ const Simulator = ({
   const simHeight = contentRect?.height ?? 0;
   const [brand500] = useToken("colors", ["brand.500"]);
   const [running, setRunning] = useState<RunningStatus>(RunningStatus.STOPPED);
+  const previouslyShown = usePrevious(shown);
 
   useEffect(() => {
     if (shown) {
-      ref.current!.focus();
+      // Prevents the simulator stealing focus on initial load.
```
microbit-matt-hillsdon, line 81: I'd really love to find a way to move away from effects for this but I think this is a reasonable fix at this point.

src/workbench/SideBarTab.tsx
```diff
@@ -29,8 +29,9 @@ const SideBarTab = ({
   const ref = useRef<HTMLButtonElement>(null);
   useEffect(() => {
     // Override tabindex.
+    // Has no dependencies as it needs to run for every re-render.
     ref.current!.setAttribute("tabindex", "0");
```
microbit-matt-hillsdon, line 33: Seems reasonable, though I'd love a world where we didn't have to hack this.

PR #983, opened 2022-09-21
Add e2e tests to check focus on load and user actions

microbit-robert:
Useful to confirm that focus is working as expected to maintain accessibility standards.

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -124,6 +124,12 @@ export class App {
       value: "1",
       url: this.url,
     });
+    // Don't show compliance notice.
+    await page.setCookie({
+      name: "MBCC",
+      value: "1",
+      url: this.url,
+    });
 
     const client = await page.target().createCDPSession();
     await client.send("Page.setDownloadBehavior", {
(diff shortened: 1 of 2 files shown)


src/e2e/app.ts
```diff
@@ -1139,6 +1145,42 @@ export class App {
...
+      query
+    );
+  }
+
+  async compareElementHandles(
+    e1: puppeteer.ElementHandle<Element>,
+    e2: puppeteer.ElementHandle<Element>
+  ): Promise<boolean> {
+    return (await this.page).evaluate((e1, e2) => e1 === e2, e1, e2);
+  }
```
microbit-matt-hillsdon, line 1182: If we expose all this then it's quite likely the e2e tests will deteriorate into a big mess with lots of details of the UI text in different test files. Please can you have a think about whether there's a way to test this that better encapsulates things in app.ts.

microbit-robert: [`5ec4103`](https://github.com/microbit-foundation/python-editor-v3/pull/983/commits/5ec4103dc0c8784bab2b06d3729c48715d61466d)

PR #991, opened 2022-09-22
Ensure focus returns to menu buttons for accessibility purposes

### LINKED ISSUE #974: Focus gets reset when exitting "Connect" dialog via "more connect options"
## Bug Description

GOOD: When you use the "Send to micro:bit" button and then cancel the "Connect cable" dialog, the focus is correctly put back on the "Send to micro:bit" button. Hurrah.

BAD: If you access the same dialog from the "more connect options" button -> "Connect" workflow, focus is reset to the top left of the screen and if you are navigating with a screen reader or keyboard control, then it is a pain to get back to the project actions. 

## How To Reproduce

1. There is no micro:bit connected
2. Click on the 'more actions' menu button
3. Click on "Connect"
 image
4. Exit the dialog either by pressing escape, clicking on the "X" or "Cancel" or using keyboard controls to select "X" or "Cancel"
5. Press tab to observe focus 

## Expected behavior

Focus should be put back to the 'more actions' menu button or at least the project actions region 

## Environment

Windows 10 / Chrome / live site

microbit-robert:
Closes #974


microbit-robert:
The amount of prop and ref passing is extreme, but perhaps unavoidable. The connection helper dialogs make this harder, but I think I've tightened up behaviour here as well. 

Todo:

- WebUSB not supported flow.
- Check firmware update flow.

Will need testing to ensure that uncommon cases haven't been missed.

PR #992, opened 2022-09-22
Style images inside documentation

microbit-robert:


src/common/imageUrlBuilder.ts
```diff
@@ -16,12 +16,12 @@ export const imageUrlBuilder = unconfiguredImageUrlBuilder()
...
 
-export const getAspectRatio = (imageRef: string): number | undefined => {
+export const getAspectRatio = (imageRef: string): string | undefined => {
   const dimensionsArr = imageRef.match(/\d+x\d+/g);
   if (!dimensionsArr) {
     return undefined;
   }
   const dimensions = dimensionsArr.join().split("x");
   const [width, height] = dimensions.map((n: string) => Number(n));
-  return width / height;
```
microbit-robert, line 26: This number was being interpreted as a pixel value which is invalid CSS. Returning a string fixes this.

microbit-matt-hillsdon: I'd noticed this wasn't working well but hadn't looked again at the code. Thanks.

PR #993, opened 2022-09-23
Add accelerometer gestures to UI strings

microbit-robert:
This PR also applies the code font family to the select gesture dropdown in the Simulator.

Preparation to translate the gestures in the same way that the API names have been. E.g. "English gesture (translated gesture)".

### DIFF
--- a/src/simulator/AccelerometerModule.tsx
+++ b/src/simulator/AccelerometerModule.tsx
@@ -113,6 +113,7 @@ const Gesture = ({ icon, state, enabled, onValueChange }: GestureProps) => {
         value={choice}
         onChange={handleSelectChange}
         ref={selectRef}
+        fontFamily="code"
       >
         {choices.map((choice) => (
           <option key={choice} value={choice}>
--- a/src/simulator/AccelerometerModule.tsx
+++ b/src/simulator/AccelerometerModule.tsx
@@ -109,7 +109,7 @@ const Gesture = ({ icon, state, enabled, onValueChange }: GestureProps) => {
       id: `simulator-gesture-${gesture.split(" ").join("-")}`,
     });
     if (translation !== gesture) {
-      gesture += ` (${translation})`;
+      return `${translation} (${gesture})`;
     }
     return gesture;
   };
@@ -123,7 +123,6 @@ const Gesture = ({ icon, state, enabled, onValueChange }: GestureProps) => {
         value={choice}
         onChange={handleSelectChange}
         ref={selectRef}
-        fontFamily="code"
       >
         {choices.map((choice) => (
           <option key={choice} value={choice}>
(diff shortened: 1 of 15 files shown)


src/simulator/AccelerometerModule.tsx
```diff
@@ -113,10 +113,13 @@ const Gesture = ({ icon, state, enabled, onValueChange }: GestureProps) => {
...
         onChange={handleSelectChange}
         ref={selectRef}
+        fontFamily="code"
       >
         {choices.map((choice) => (
           <option key={choice} value={choice}>
-            {choice}
+            {intl.formatMessage({
+              id: `simulator-gesture-${choice.split(" ").join("-")}`,
+            })}
```
microbit-matt-hillsdon, line 122: I guess you could make this work for the final state now by showing it bracketed if it's different from the English? I think that's what we intend, though not sure which way round. Please can you follow up on that discussion and check there's consensus?

microbit-robert: Unable to render any HTML in the option tag, so the font family will be the same -> [`4fc19e8`](https://github.com/microbit-foundation/python-editor-v3/pull/993/commits/4fc19e859dee39451d9a7fee4c674ee2c4ab5e27).

src/simulator/AccelerometerModule.tsx
```diff
@@ -104,6 +104,16 @@ const Gesture = ({ icon, state, enabled, onValueChange }: GestureProps) => {
     }, 500);
   }, [setActive, onValueChange, choice]);
 
+  const renderOptionText = (gesture: string): string => {
+    const translation = intl.formatMessage({
+      id: `simulator-gesture-${gesture.split(" ").join("-")}`,
+    });
+    if (translation !== gesture) {
+      gesture += ` (${translation})`;
```
microbit-matt-hillsdon, line 112: Agreed this morning that we'd put the translation first.

src/simulator/AccelerometerModule.tsx
```diff
@@ -113,10 +123,11 @@ const Gesture = ({ icon, state, enabled, onValueChange }: GestureProps) => {
         value={choice}
         onChange={handleSelectChange}
         ref={selectRef}
+        fontFamily="code"
```
microbit-matt-hillsdon, line 126: Agreed this morning that we'd remove this as we can't apply it only to the English text and it doesn't make sense for the translation.

src/simulator/AccelerometerModule.tsx
microbit-robert: [`f6037ec`](https://github.com/microbit-foundation/python-editor-v3/pull/993/commits/f6037ecd0b16471f40d369e3c9e515d5065b4660)

src/simulator/AccelerometerModule.tsx
microbit-robert: [`f6037ec`](https://github.com/microbit-foundation/python-editor-v3/pull/993/commits/f6037ecd0b16471f40d369e3c9e515d5065b4660)

microbit-matt-hillsdon (approved):
LGTM. I'll do a translation sync late today or tomorrow so the strings are available for translation.

microbit-matt-hillsdon:
Will merge once it's rebuilt.

PR #997, opened 2022-09-26
Use optional chaining to prevent button reference error.

microbit-robert:
This is a logged bug that appears to occur after sending a gesture to the simulator and then expand/collapsing the accelerometer module panel. This causes a re-render and the gesture button ref can be undefined. We should also not change the focus if the user's last action was to expand/collapse the module, so this seems a reasonable fix.


microbit-matt-hillsdon:
We can definitely merge this, but perhaps we shouldn't disable the button at all? E.g. a quick enter, tab is quite confusing. I'm assuming disabling the button is what's preventing focus staying on it. Feels like it needs some kind of reaction though.

microbit-matt-hillsdon (approved):

PR #999, opened 2022-09-26
Align social meta with live except image.

microbit-matt-hillsdon:
Image matches the update made to the versioner.


public/index.html
```diff
@@ -11,7 +11,7 @@
       property="og:description"
       content="Built by the Micro:bit Educational Foundation and the global Python Community."
     />
-    <meta property="og:image" content="%PUBLIC_URL%/social.png"/>
+    <meta property="og:image" content="https://python.microbit.org/%PUBLIC_URL%social.png"/>
```
microbit-robert, line 14: Missing slash between %PUBLIC_URL% and social?

microbit-matt-hillsdon: It really does seem unreasonable that og:image has to be absolute.

PR #1002, opened 2022-09-26
Add heading elements to dialog headers to improve accessibility

### LINKED ISSUE #998: "Confirm replace project" dialog has no headings
## Bug Description
The dialog you get when opening an example has a `header` but no `h1` / `h2` headings which means a screen reader user can't skip to the title with header navigation. 

image

[edit: confusing typo heading/header]

microbit-robert:
Closes #998.

### DIFF
--- a/src/common/GenericDialog.tsx
+++ b/src/common/GenericDialog.tsx
@@ -48,8 +48,8 @@ export const GenericDialog = ({
         <ModalContent minWidth="560px" my="auto">
           <ModalCloseButton />
           {header && (
-            <ModalHeader fontSize="lg" fontWeight="bold">
-              <Text as="h2" fontSize="xl">
+            <ModalHeader>
+              <Text as="h2" fontSize="xl" fontWeight="bold">
                 {header}
               </Text>
             </ModalHeader>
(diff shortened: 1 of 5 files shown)


microbit-matt-hillsdon:
Looks like a related e2e failure.

src/common/GenericDialog.tsx
```diff
@@ -49,7 +49,9 @@ export const GenericDialog = ({
           <ModalCloseButton />
           {header && (
             <ModalHeader fontSize="lg" fontWeight="bold">
-              {header}
+              <Text as="h2" fontSize="xl">
```
microbit-matt-hillsdon, line 52: Let's sort the inconsistency here, though as discussed this code is not used.

microbit-robert: Made consistent with other uses of the `GenericDialog` component where the heading is included in the dialog body -> [`2a00984`](https://github.com/microbit-foundation/python-editor-v3/pull/1002/commits/2a00984455398ad152a33cf41e9970990fbd825d). 

Use of the header prop changes the appearance significantly, so is not used for the existing connect helper dialogs.

PR #1004, opened 2022-09-27
Enable explanatory dialogs for all languages.

microbit-matt-hillsdon:
These were disabled due to lack of translated text.


microbit-robert (approved):
LGTM. Dialogs now appear in other languages now translations are complete.

PR #1007, opened 2022-09-30
Updates for when this is the live editor.

microbit-matt-hillsdon:


jaustin (approved):
These changes look great. 

Once we get past the immediate launch, I suggest we add a section on accessibility, and something on how we plan to manage development and research people might be interested in doing on the editor.

PR #1009, opened 2022-09-30
Avoid unwanted project persistence

microbit-matt-hillsdon:
Session storage lasts for the lifetime of the tab, so we need to make sure subsequent migration URLs and iframe configurations work.

Motivated by issues with "Open in Classroom" and "Open in Python" on microbit.org.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "@microbit-foundation/python-editor-v3",
+  "name": "@microbit/python-editor-v3",
   "version": "3.0.0-local",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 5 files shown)


src/fs/host.test.ts
```diff
@@ -23,7 +23,7 @@ describe("IframeHost", () => {
   delete (window as any).parent;
   (window as any).parent = parentWindow;
   delete (window as any).location;
-  window.location = new URL("https://localhost:3000?controller=1") as any;
+  window.location = new URL("https://localhost?controller=1") as any;
```
microbit-matt-hillsdon, line 26: Not sure I needed to change this at all.

jaustin:
I can switch between the two code snippets by navigating back/forward in this tab from the URLs you've provided (Ie this appears to fix the issue for me, here)

PR #1017, opened 2022-10-05
Fix for simulated button clicks for simulator button/pin inputs

### LINKED ISSUE #984: [Simulator] Button presses are sometimes holds
## Bug Description

In the simulator you can press button A or B. By default this is supposed to be a "press and release" unless you expand to access the "hold" toggle switches. 

However, some methods of clicking the buttons leave them incorrectly stuck down. So far I've found touch input, and one particular type or simulated mouse click. But I imagine it's a more general issue to do with the timings/specifics of mouse up/down/press events.

## How To Reproduce

Simplest way to reproduce is with a touch screen device:
1. Run the sample code in the simulator
2. Tap button A or B with your finger
  - expected behaviour: button is pressed and then released
  - observed behaviour: button is held down
3. Tap again and button is released.
4. Tap again and button is held down
5. Now click the held-down button with your mouse
  - The button now works in reverse, briefly being released while you click it and then being pressed again

The same scenario can be reproduced using [Optikey](https://github.com/OptiKey/OptiKey/) to simulate mouse presses with eye control.

## Environment

Windows 10/Chrome/Surface Pro touchscreen

microbit-robert:
Closes #984.

### DIFF
--- a/src/simulator/SensorInput.tsx
+++ b/src/simulator/SensorInput.tsx
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  */
 import { Box, Button, Switch, VStack } from "@chakra-ui/react";
-import { ReactNode, useCallback, useState } from "react";
+import { ReactNode, useCallback, useRef, useState } from "react";
 import { useIntl } from "react-intl";
 import {
   RangeSensor as RangeSensorType,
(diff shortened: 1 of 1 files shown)


microbit-robert:
The simulated click seemingly has no delay between mousedown and mouseup which is causing us to end up with unpredictable state in an already complex component. Switching from using state to refs seems to fix the issue, though I'm having trouble reasoning about this change and haven't explored the potential for any unwanted side effects of doing so.

microbit-matt-hillsdon:
Code-wise this makes sense to me. We only use mouseDown in event handlers and need it to be up-to-date (can't wait for a re-render). I'll play with it some more then approve.

microbit-matt-hillsdon:
Review link (not sure why missing):

https://review-python-editor-v3.microbit.org/optikey-fix/

microbit-matt-hillsdon (approved):
Working well. @microbit-robert this is good so far as I'm concerned. I moved it out of draft so wanted to check you were happy before merging. If so feel free to go ahead.

PR #1020, opened 2022-10-06
Format message for documentation navigation event

microbit-robert:


src/router-hooks.tsx
```diff
@@ -112,9 +112,11 @@ export const RouterProvider = ({ children }: { children: ReactNode }) => {
   const navigate = useCallback(
     (newState: RouterState, source?: NavigationSource) => {
       if (source) {
+        const parts = toUrl(newState).split("/");
+        const message = parts.slice(parts.length - 2).join("-");
```
microbit-matt-hillsdon, line 116: I suppose I'd like to do this in a way that doesn't rely on the deployment path (e.g. now it'll be weird for local dev). Can we extract something from toUrl and use that? Or strip off the known prefix?

microbit-matt-hillsdon (approved):
LGTM.

Issue #1006, opened 2022-09-30
Capitalisation issue in French translation


microbit-matt-hillsdon:
The data logging string shown here has no initial capital letter. Not a big issue, but it's shown by default so would be nice to fix.

Fix needs to be in Crowdin then a translation sync.

![image](https://user-images.githubusercontent.com/44397098/193223560-6bb4a150-f8e4-4528-a506-4256dcd987f1.png)

microbit-matt-hillsdon:
Corrected in Crowdin here: https://crowdin.com/translate/microbitorg/6426/en-fr#334686

I'll do another translation sync after we've had a chance for more translation feedback from the live release and this will be included in that.

Closing the issue.

microbit-matt-hillsdon:
Also checked Spanish and it was OK there.

PR #1029, opened 2022-10-12
Update translations.

microbit-matt-hillsdon:
No pyright changes this time around.
Script updated to move some of the logic into the stubs repo.


lang/ui.fr.json
```diff
@@ -827,7 +827,7 @@
     "description": "Shown below the simulator Data logging table to warn that the log is full"
   },
   "simulator-data-logging-rows": {
-    "defaultMessage": "{count, plural, =0 {aucune ligne enregistrée} one {# ligne enregistrée} other {# lignes enregistrées}}",
```
microbit-matt-hillsdon, line 830: This is the fix for #1006

lang/ui.fr.json
```diff
@@ -859,15 +859,15 @@
     "description": "Simulator gesture option"
   },
   "simulator-gesture-down": {
-    "defaultMessage": "logo en bas",
```
microbit-matt-hillsdon, line 862: This is expected change as we used some not-quite-right translations from Reference as a stop gap.

PR #1031, opened 2022-10-14
Relax create file lowercase validation check

microbit-matt-hillsdon:
See https://github.com/microbit-foundation/python-editor-v3/issues/1026


microbit-matt-hillsdon:
Code change looks ok. Needs careful review of:

- [ ] Create file dialog
- [ ] Project name dialog
- [ ] File confirmation dialog (though should be no impact due to lack of validation)

These all share the common input dialog code that now allows for validation messages that aren't form errors.

PR #1036, opened 2022-10-17
Correct Dutch language name

microbit-matt-hillsdon:


src/settings/settings.tsx
```diff
@@ -61,6 +61,7 @@ export const supportedLanguages: Language[] = [
     enName: "Spanish",
   },
 ].filter((l) => stage !== "PRODUCTION" || !l.preview);
+console.log(stage, supportedLanguages);
```
microbit-matt-hillsdon, line 64: Remove debug.

```suggestion
```

PR #1037, opened 2022-10-18
Drag and drop fixes

microbit-matt-hillsdon:
- Default to appending to existing blocks but allow control over the indent level where it can't cause invalid indentation.
- Fix handling of whitespace in "while smooshing"

Closes https://github.com/microbit-foundation/python-editor-v3/issues/856.
Closes https://github.com/microbit-foundation/python-editor-v3/issues/812.
Closes https://github.com/microbit-foundation/python-editor-v3/issues/800.
Closes https://github.com/microbit-foundation/python-editor-v3/issues/539.

Concerns:
- As we don't detect intent level this weds us further to 4 space indents. Easily generalised if we did autodetect or have a preference in future. I think if you do either you need to do both though. Code structure highlighting is enabled by default and looks very poor if you don't follow the convention so I think this is OK. Likewise the autoindent and tab behaviours.
- Build change to allow iterating over generators. Needs a quick check in supported browsers but really should be fine.

### DIFF
--- a/src/editor/codemirror/copypaste.ts
+++ b/src/editor/codemirror/copypaste.ts
@@ -51,9 +51,9 @@ const copyPasteHandlers = () => [
         message: pasteContext.id,
       });
 
-      const lineNumber = view.state.doc.lineAt(
-        view.state.selection.ranges[0].from
-      ).number;
+      const line = view.state.doc.lineAt(view.state.selection.ranges[0].from);
+      const lineNumber = line.number;
+      const column = view.state.selection.ranges[0].from - line.from;
 
       view.dispatch(
         calculateChanges(
@@ -61,6 +61,7 @@ const copyPasteHandlers = () => [
           pasteContext.codeWithImports,
           pasteContext.type,
           lineNumber,
+          Math.floor(column / 4),
           true
         )
       );
--- a/src/editor/codemirror/edits.ts
+++ b/src/editor/codemirror/edits.ts
@@ -112,6 +112,21 @@ export const calculateChanges = (
         } else {
           mainFrom = state.doc.line(line).from;
         }
+        // Determine if code is being appended to previous block and indent accordingly.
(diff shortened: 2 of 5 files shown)


src/editor/codemirror/copypaste.ts
```diff
@@ -51,16 +51,17 @@ const copyPasteHandlers = () => [
...
+      const lineNumber = line.number;
+      const column = view.state.selection.ranges[0].from - line.from;
 
       view.dispatch(
         calculateChanges(
           view.state,
           pasteContext.codeWithImports,
           pasteContext.type,
           lineNumber,
+          Math.floor(column / 4),
```
microbit-matt-hillsdon, line 64: I've not tried out the experience with copy/paste.

src/editor/codemirror/edits.ts
```diff
@@ -153,6 +156,61 @@ export const calculateChanges = (
...
+    const text = line.text;
+    if (text.trim()) {
+      let indent = indentLevel(text);
+      // Beginning of block:
+      if (text.trim().endsWith(":")) {
+        indent += 1;
+      } else if (
+        levelHint !== undefined &&
+        levelHint < indent &&
+        nextNonBlankLineIndentLevel < indent &&
```
microbit-matt-hillsdon, line 185: We also shouldn't be able to go back further than nextNonBlankLineIndentLevel. So we need the max of that and the hint.

microbit-matt-hillsdon: I didn't express is quite right above. Clearer in code: e865038d866e5eb386117e01fe03422e889ac033

microbit-matt-hillsdon:
> Build change to allow iterating over generators. Needs a quick check in supported browsers but really should be fine.

This looks good. I don't think it's really changed anything other than the input to babel.

PR #1047, opened 2022-10-23
Update translations.

microbit-matt-hillsdon:
Stubs changes on https://github.com/microbit-foundation/micropython-microbit-stubs/pull/83


microbit-matt-hillsdon:
Changes look good. Affects Dutch only which is just live on the beta at this point.

PR #1049, opened 2022-10-24
Reinstate keyboard escape hatch for editor

### LINKED ISSUE #1048: Escape, Tab to tab out of the editor is broken
Escape, Tab to tab out of the editor is broken. We suspect the CodeMirror 6 upgrade. As it's otherwise not possible to tab out we should fix this quickly.

microbit-robert:
Closes #1048.


microbit-matt-hillsdon (approved):
LGTM

PR #1051, opened 2022-10-24
Test tab out of editor scenarios

microbit-matt-hillsdon:
Try to make the existing focus tests simpler to follow. 
Create a method to assert focus that we can use to create more such assertions.


microbit-matt-hillsdon:
Latest failure was in the CF invalidation. Re-running.

microbit-matt-hillsdon:
Passed this time but queuing again for confirmation as I hit various flakiness when writing (though not with the latest code).

microbit-matt-hillsdon:
> Passed this time but queuing again for confirmation as I hit various flakiness when writing (though not with the latest code).

All looks good.

microbit-robert (approved):
LGTM.

Issue #1053, opened 2022-10-27
[CodeMirror] Invalid position errors due to stale tooltips


microbit-matt-hillsdon:
We see common traces like this:

RangeError
Invalid position 125 in document of length 122 __numbers vary though this one is most common__
```
Crashed in non-app: ../node_modules/@codemirror/state/dist/index.js in lineAt
<anonymous> in Array.map
Called from: ../node_modules/@codemirror/view/dist/index.js in read
<anonymous> in Array.map
Called from: ../node_modules/@codemirror/view/dist/index.js in measure
../../../../browser/src/helpers.ts in i at line 73:23
```

This amounts to a crash in the tooltip module.

I've managed to reproduce it by inserting a significant delay in the removal of a signature help tooltip on the last line of the document and then deleting the line.

We should review all asynchronous tooltip handling.

Detailed repro:

- Add `await new Promise((resolve) => setTimeout(resolve, 500));` to the catch block in triggerSignatureHelpRequest (signatureHelp.ts)
- Use this file with no trailing line

```python
# Imports go at the top
from microbit import *


# Code in a 'while True:' loop repeats forever
while True:
    display.show(Image.HEART)
    sleep(1000)
    display.scroll()
```

- Trigger signature help in the brackets with Mod-Shift-Space
- Delete the line with Mod-Shift-K

I suspect there are other repros, but the general idea of it being a tooltip for an invalid position due to async is likely common to them.

Pretty trace from a repro:

```
react_devtools_backend.js:4026 RangeError: Invalid position 173 in document of length 153
    at TextLeaf.lineAt (index.js:13:1)
    at EditorView.coordsAtPos (index.js:6678:1)
    at index.js:8203:1
    at Array.map (<anonymous>)
    at Object.readMeasure (index.js:8201:1)
    at index.js:6380:1
    at Array.map (<anonymous>)
    at EditorView.measure (index.js:6378:1)
    at index.js:6504:1
```

microbit-matt-hillsdon:
lint and signature help have this characteristic, but lint maps errors through document changes so I guess would move the error to the end of the document (and then it'd be replaced by a diagnostics update). So I think we need to do something similar with the position we're intending to show signature help for.

Note: 125, which is the common position for this error is the cursor position in the brackets of a display.show in the default while loop, so that makes a lot of sense.

microbit-matt-hillsdon:
[...]

PR #1054, opened 2022-10-27
Fix RangeError due to stale signature help tooltips

microbit-matt-hillsdon:
It now maps the position through so you see lagged tooltips if Pyright responses are slow rather than tooltips for positions that might not be in the document at all.

### DIFF
--- a/src/editor/codemirror/language-server/signatureHelp.ts
+++ b/src/editor/codemirror/language-server/signatureHelp.ts
@@ -43,7 +43,7 @@ export const setSignatureHelpResult = StateEffect.define<SignatureHelp | null>(
   {}
 );
 
-interface SignatureHelpState {
+class SignatureHelpState {
   /**
    * -1 for no signature help requested.
    */
@@ -55,6 +55,14 @@ interface SignatureHelpState {
    * but we display it as it's generally useful.
    */
   result: SignatureHelp | null;
+
+  constructor(pos: number, result: SignatureHelp | null) {
+    if (result && pos === -1) {
+      throw new Error("Invalid state");
+    }
+    this.pos = pos;
+    this.result = result;
+  }
 }
 
 const signatureHelpToolTipBaseTheme = EditorView.baseTheme({
@@ -113,10 +121,7 @@ export const signatureHelp = (
   apiReferenceMap: ApiReferenceMap
 ) => {
   const signatureHelpTooltipField = StateField.define<SignatureHelpState>({
-    create: () => ({
-      pos: -1,
-      result: null,
-    }),
+    create: () => new SignatureHelpState(-1, null),
     update(state, tr) {
       let { pos, result } = state;
       for (const effect of tr.effects) {
@@ -140,10 +145,7 @@ export const signatureHelp = (
         // Avoid pointless tooltip updates. If nothing else it makes e2e tests hard.
         return state;
       }
-      return {
-        pos,
-        result,
-      };
+      return new SignatureHelpState(pos, result);
     },
     provide: (f) =>
       showTooltip.from(f, (val) => {
--- a/src/editor/codemirror/language-server/signatureHelp.ts
+++ b/src/editor/codemirror/language-server/signatureHelp.ts
@@ -134,7 +134,6 @@ export const signatureHelp = (
     provide: (f) =>
       showTooltip.from(f, (val) => {
         const { result, pos } = val;
-        console.log(result, pos);
         if (result) {
           return {
             pos,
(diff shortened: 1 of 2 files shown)


src/editor/codemirror/language-server/signatureHelp.ts
```diff
@@ -99,27 +109,55 @@ export const signatureHelp = (
 ) => {
   const signatureHelpTooltipField = StateField.define<SignatureHelpState>({
     create: () => ({
+      pos: -1,
       result: null,
-      tooltip: null,
     }),
     update(state, tr) {
+      let pos = state.pos === -1 ? -1 : tr.changes.mapPos(state.pos);
+      console.log(state.pos, pos);
```
microbit-matt-hillsdon, line 117: Remove debug.

```suggestion
```

src/editor/codemirror/language-server/signatureHelp.ts
```diff
@@ -99,27 +109,55 @@ export const signatureHelp = (
...
+      return {
+        pos,
+        result,
+      };
     },
-    provide: (f) => showTooltip.from(f, (val) => val.tooltip),
+    provide: (f) =>
+      showTooltip.from(f, (val) => {
+        const { result, pos } = val;
+        console.log(result, pos);
```
microbit-matt-hillsdon, line 138: Remove debug.

```suggestion
        console.log(result, pos);
```

microbit-matt-hillsdon: You see this as (125, 124) in the scenario that otherwise would be bug #1053

microbit-matt-hillsdon: Not sure why I can't commit this from the GH web UI. Will resolve and re-raise.

microbit-matt-hillsdon: Ah, I see, it was a no-op change. Edited to remove.

src/editor/codemirror/language-server/signatureHelp.ts
```diff
@@ -72,17 +77,22 @@ const triggerSignatureHelpRequest = async (view: EditorView): Promise<void> => {
...
     view.dispatch({
-      effects: [setSignatureHelpEffect.of({ pos, result })],
+      effects: [setSignatureHelpResult.of(result)],
     });
   } catch (e) {
     logException(view.state, e, "signature-help");
     view.dispatch({
-      effects: [setSignatureHelpEffect.of({ pos, result: null })],
+      effects: [setSignatureHelpResult.of(null)],
     });
```
microbit-matt-hillsdon, line 96: The essence of the change is that we no longer assume this position is still valid when we hear back from Pyright.
Instead we stash it before the request and map it forward through subsequent changes as required.

microbit-matt-hillsdon:
> The slightly buggy behaviour of having signature help on an existing function open, and then dragging over a new API snippet I mentioned previously causes errors on this branch.

Thanks, I can reproduce them.

src/editor/codemirror/language-server/signatureHelp.ts
```diff
@@ -63,32 +68,42 @@ const signatureHelpToolTipBaseTheme = EditorView.baseTheme({
...
+  const client = state.facet(clientFacet)!;
+  const pos = state.selection.main.from;
   const params: SignatureHelpParams = {
     textDocument: { uri },
-    position: offsetToPosition(view.state.doc, pos),
+    position: offsetToPosition(state.doc, pos),
   };
   try {
+    // Must happen before other event handling that might dispatch more
+    // changes thatt invalidate our position.
```
microbit-matt-hillsdon, line 84:
```suggestion
    // changes that invalidate our position.
```

microbit-matt-hillsdon:
I think this is now working. @microbit-robert, please break it again!

I want this to work without the fix of closing signature help as I think dnd actually exercises this code really well. However, that's still a good plan so I've raised #1055 and have a fix on #1056 that I'll rebase after this is in.

microbit-matt-hillsdon:
I'm re-running CI (it passed) to build a bit more confidence in the e2e tests here.

microbit-matt-hillsdon:
Looking again at this as I managed to trigger an error with my original repro.

microbit-matt-hillsdon:
> Looking again at this as I managed to trigger an error with my original repro.

OK, I'm happy again 🤞

It's rebuilding now.

microbit-robert (approved):
Now looks good. I can occasionally see some lag in updating the tooltip as expected, but I can't produce any errors.

PR #1056, opened 2022-10-28
Close signature help on blur

microbit-matt-hillsdon:
Depends on / needs rebasing after #1054


microbit-matt-hillsdon:
Hah, the blur breaks interaction with Help/API links!

microbit-matt-hillsdon:
> Hah, the blur breaks interaction with Help/API links!

Since fixed & rebased.

Issue #1060, opened 2022-11-02
RangeError in structure highlighting


microbit-matt-hillsdon:
```
RangeError: Invalid position 978 in document of length 977
  at lineAt(../node_modules/@codemirror/state/dist/index.js:14:19)
  at lineAt(../node_modules/@codemirror/view/dist/index.js:4327:44)
  at lineAt(../node_modules/@codemirror/view/dist/index.js:4422:26)
  at lineAt(../node_modules/@codemirror/view/dist/index.js:4422:26)
  at lineAt(../node_modules/@codemirror/view/dist/index.js:4422:26)
  at lineAt(../node_modules/@codemirror/view/dist/index.js:4422:26)
  at lineAt(../node_modules/@codemirror/view/dist/index.js:4422:26)
  at lineAt(../node_modules/@codemirror/view/dist/index.js:4422:26)
  at lineBlockAt(../node_modules/@codemirror/view/dist/index.js:5086:39)
  at lineBlockAt(../node_modules/@codemirror/view/dist/index.js:6580:31)
  at positionsForNode(editor/codemirror/structure-highlighting/view.ts:102:28)
  at leave(editor/codemirror/structure-highlighting/view.ts:184:43)
  at iterate(../node_modules/@lezer/common/dist/index.js:355:21)
  at read(editor/codemirror/structure-highlighting/view.ts:155:16)
  at ? (../node_modules/@codemirror/view/dist/index.js:6380:34)
  at Array.map(<anonymous>)
  at measure(../node_modules/@codemirror/view/dist/index.js:6378:42)
  at onScrollChanged(../node_modules/@codemirror/view/dist/index.js:6152:22)
  at wrappedArguments(../node_modules/@codemirror/view/dist/index.js:5548:14)
  at HTMLDivElement.i(../../../../browser/src/helpers.ts:73:23)
```

Now clearer in logging since #1054. This one is unrelated.

microbit-matt-hillsdon:
The code is adding 1 to a position so it seems likely that some straightforward defensiveness would fix. Would be nice to have a repro though.

microbit-matt-hillsdon:
So we're finding the lineBlock for the start of the body. I think that starts with the colon.
But visually we want to highlight the next line, so we're moving to the next line via the +1.
But that position might not exist.
So far as I can tell it routinely does not exist in scenarios like

```python
if True: pass
```

...but lineBlockAt doesn't fail in in these obvious reproductions.

[...]

PR #1061, opened 2022-11-02
Structure highlighting: don't ask for line information past EOF

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-v3/issues/1060

Can be reproduced with `if True:pass` at EOF and, importantly, outside the viewport (so prepended with blank lines, for example).

Also changes the highlighting in this case, see before/after images on #1060.

### DIFF
--- a/src/editor/codemirror/structure-highlighting/view.ts
+++ b/src/editor/codemirror/structure-highlighting/view.ts
@@ -86,24 +86,36 @@ export const codeStructureView = (option: "full" | "simple") =>
 
         let cursorFound = false;
 
+        /**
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
I think this fix is fine but I'd like to look more carefully at what we do with single line bodies.
There's code that looks like it special cases them but I don't think it's functioning as intended.

microbit-matt-hillsdon:
> I think this fix is fine but I'd like to look more carefully at what we do with single line bodies. There's code that looks like it special cases them but I don't think it's functioning as intended.

This is now addressed.

PR #1064, opened 2022-11-04
Fix bug if lines don't match when mapping diagnostics

microbit-matt-hillsdon:
The document may have changed in the meantime so notice when lines don't make sense. I've also added a specific check for the scenario where diagnostics have to < from.

Closes https://github.com/microbit-foundation/python-editor-v3/issues/1062


microbit-matt-hillsdon:
Tried point and range diagnostics and all seems to be working well. I don't think it's reasonable to be able to repro this via the UI so we'll have to rely on reviewing the logic of the fix and the tests.

microbit-robert (approved):
LGTM.

microbit-matt-hillsdon:
I'll monitor the impact on staging for a bit before releasing this.

PR #1066, opened 2022-11-04
Fix early DOM access

microbit-matt-hillsdon:
I'm not sure if this fix makes sense. In production the main script has the defer attribute which is supposed to wait until after the parse. But I'm not sure what else explains the crashes we see here. Leaving here as a draft to ponder.


microbit-matt-hillsdon:
I also can't repro document.readyState < interactive in a minimal setup (even with a large DOM). I wonder what's going on here.

microbit-matt-hillsdon:
I think this is based on a misunderstanding. There must be another explanation. Closing.

PR #1067, opened 2022-11-05
Signature help fix tweak.

microbit-matt-hillsdon:
Close the signature help while we know the view is still good. Alternatively we could cancel the timeout but this is OK too. I can't see simple API to check that a view is still good to interact with after an async step.


microbit-matt-hillsdon:
We've seen crashes where this dispatch results in CM accessing a contentEditable field that's no longer there. I assume some blur scenario that results in a change of editor but I wasn't able to repro even with a manually added timeout.

PR #1074, opened 2022-11-16
Audit fixes, including craco 7.0.0.

microbit-matt-hillsdon:
Covers current dependabot alert.


microbit-matt-hillsdon:
Audit fix upgraded Craco in package-lock only but I moved it to package.json for clarity.
A bit of prodding would be great. I've done some testing in Safari.

microbit-robert (approved):
All seems well. The theme package and search indexing, which is what we use craco for, are still working without issue.

PR #1083, opened 2022-12-15
[DEMO - DO NOT MERGE] Demo opening hex directly in the micro:bit app

microbit-robert:


microbit-robert:
Does not check what browser/device you are using. Just assumes you are on iPad/mobile and attempts to open the project hex in the micro:bit app. Not sure what happens if you don't have the app downloaded, but it is likely to fail in the same semi-silent way that MakeCode does.

MakeCode manages to preserve project titles with spaces, this implementation resorts to URI encoding. Needs fixing.

martinwork:
MakeCode shows a banner if the app is not installed.

microbit-robert:
Yep, MakeCode in Safari shows a banner at the top prompting the user to either install the app, or open in the app. 

The Python Editor V3 doesn't currently support opening a project as a base64 encoded .hex file appended on the URL. We'd need to change this first before being able to launch via the app.

martinwork:
Right. sorry, I was meaning... is that the intended method?

PR #1089, opened 2023-01-18
Enable Dutch for production

microbit-matt-hillsdon:
It has been proofread and approved by Emma.


microbit-matt-hillsdon:
Now in the list just like any other language:

image

I've kept the preview logic with minor faff to avoid a type inference issue.

microbit-robert (approved):
LGTM.

PR #1096, opened 2023-01-26
Actions: store reports on failure not success only

microbit-matt-hillsdon:


microbit-robert (approved):
LGTM. I'm assuming the e2e test failure is due to some flakiness(?) as nothing else has changed.

microbit-matt-hillsdon:
> LGTM. I'm assuming the e2e test failure is due to some flakiness(?) as nothing else has changed.

Yes, though we need to prioritise looking at this as it's significantly worse since the move to GitHub Actions.

PR #1097, opened 2023-01-30
Route the measurement id through workflow

microbit-matt-hillsdon:


.github/workflows/build.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       - run: node ./bin/print-ci-env-public-url.js >> $GITHUB_ENV
       - run: npm run ci:update-version
       - run: npm run ci
+        env:
+          REACT_APP_GA_MEASUREMENT_ID: ${{ secrets.REACT_APP_GA_MEASUREMENT_ID }}
```
microbit-matt-hillsdon, line 47: I think we should lose the REACT_APP_ prefix from the actual secret name.

```suggestion
          REACT_APP_GA_MEASUREMENT_ID: ${{ secrets.GA_MEASUREMENT_ID }}
```

microbit-robert (approved):
LGTM.

PR #1098, opened 2023-01-30
e2e tweaks to help debug

microbit-matt-hillsdon:
Aligns with other Foundation projects.
Only one minor fix in this PR, we definitely need more.


microbit-robert (approved):
LGTM.

PR #1103, opened 2023-02-17
WebUSB: On serial connection check baud rate before setting it.

microbit-carlos:
Setting the baud rate makes DAPLink send a "break signal", which basically resets the target.

This is noticeable to the Python Editor users when switching tabs as it makes their programme reset.

We can avoid this reset by checking first the DAPLink serial settings and only set the baud rate if necessary.

For devices with DAPLink 0257 and newer (V2.2) the default baud rate is set to 115200 already, so they will never reset on WebUSB connection. For devices with older DAPLink versions (V1 and V2.00) the editor will set the baud rate on first connection (triggering a reset), but won't have to set it again until the device is unplugged.

I tried to replicate the `Bad response for 8 -> 17` issue for about 5 minutes changing tabs and refreshing-and-reconnecting and it seems fine (and also checked that it's really easy to replicate in the Python Editor v2).
- https://github.com/microbit-foundation/python-editor-v3/issues/89

### DIFF
--- a/src/device/dap-wrapper.ts
+++ b/src/device/dap-wrapper.ts
@@ -112,7 +112,11 @@ export class DAPWrapper {
   }
 
   async startSerial(listener: (data: string) => void): Promise<void> {
-    await this.daplink.setSerialBaudrate(115200);
+    const currentBaud = await this.daplink.getSerialBaudrate();
+    if (currentBaud !== 115200) {
+      // Changing the baud rate causes a micro:bit reset, so only do it if necessary
+      await this.daplink.setSerialBaudrate(115200);
+    }
     this.daplink.on(DAPLink.EVENT_SERIAL_DATA, listener);
     await this.daplink.startSerialRead(1);
   }


src/device/dap-wrapper.ts
```diff
@@ -112,7 +112,10 @@ export class DAPWrapper {
   }
 
   async startSerial(listener: (data: string) => void): Promise<void> {
-    await this.daplink.setSerialBaudrate(115200);
+    const currentBaud = await this.daplink.getSerialBaudrate();
+    if (currentBaud != 115200) {
```
microbit-robert, line 116: Please use JavaScript's strict equality operator.
```suggestion
    if (currentBaud !== 115200) {
```

src/device/dap-wrapper.ts
```diff
@@ -112,7 +112,10 @@ export class DAPWrapper {
   }
 
   async startSerial(listener: (data: string) => void): Promise<void> {
-    await this.daplink.setSerialBaudrate(115200);
+    const currentBaud = await this.daplink.getSerialBaudrate();
```
microbit-robert, line 115: Please could you add a comment to summarize the required change here?

microbit-robert (changes requested):
This looks good to me. Tested with v1, v2.00, and v2.21, and  the program no longer resets on switching tabs. Some minor changes requested via comments.

src/device/dap-wrapper.ts
```diff
@@ -112,7 +112,10 @@ export class DAPWrapper {
...
+    if (currentBaud !== 115200) {
+      await this.daplink.setSerialBaudrate(115200);
```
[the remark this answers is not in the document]
microbit-robert: Looks good, thanks.

microbit-carlos:
Awesome, thanks @microbit-robert!
I've left the GitHub web interface commits assuming this PR will be squashed when merged, but let me know if you prefer the branch to be squashed beforehand instead.

microbit-carlos:
Ah, it needs to be rebased anyway, I'll do that

microbit-carlos:
Okay, should be ready for merge now 👍

microbit-matt-hillsdon:
Now we've done this it might be worth trying to avoid resetting the serial console (i.e clearing scrollback) in this scenario as that's what most folks have been complaining about. Needs some thought though as serial output that doesn't match the running program is confusing (e.g. we might show a clickable traceback for it).

PR #1108, opened 2023-03-03
Fix type error on Node 18/NPM 9.x

microbit-matt-hillsdon:
I think this should always have been an issue. Not clear why it wasn't previously. Type issue only, not a problem in practice.


microbit-matt-hillsdon:
e2e is the spurious toast issue we've seen often since the GHA switch. I'll spend some time on it next week. I still don't see how it can happen logically, so we'll have to catch it in the act.

microbit-robert (approved):
LGTM.

PR #1109, opened 2023-03-03
Audit fixes

microbit-matt-hillsdon:
Remainder are dev-time only and need a react-scripts update that doesn't exist and won't be prioritised by them.


microbit-matt-hillsdon:
> This version of the Python Editor doesn't currently support adding .mpy files.

There's a real consistency to these failures at least!

PR #1111, opened 2023-03-03
More reliable alert check

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-v3/issues/1110


microbit-matt-hillsdon:
As I couldn't really reproduce locally we need to see this pass in CI a few times before merge.
I still occasionally see an unrelated failure in documentation.test.ts, but at least the open.test.ts cases should pass.

PR #1113, opened 2023-03-04
More generous alert check.

microbit-matt-hillsdon:
Aims not to care about duplicates or other old alerts.


microbit-matt-hillsdon:
Failure is an unrelated (and new to me) e2e test. I'll prioritise looking at that if it recurs.

PR #1115, opened 2023-03-14
Example of a portal-based React CodeMirror widget

microbit-matt-hillsdon:
Draft PR just for ease of discussion, no intention to merge this without real usage!


microbit-matt-hillsdon:
There's a version that avoids discarding React state that could be rescued from the experiment-inline-widgets branch. Closing this PR.

PR #1116, opened 2023-03-29
Tweak based on support feedback

microbit-matt-hillsdon:
This one probably isn't worth retranslation any time soon, so I'll choose to retain the existing text when I update the content in Crowdin.


microbit-robert (approved):
LGTM.

PR #1117, opened 2023-03-31
Avoid resetting serial terminal when possible

microbit-matt-hillsdon:
Follow up to ca1f325b23cd5cbf5451cca22e94f3f245a73f3f

This aims to avoid EVENT_RESET when its due to the workaround we have in place that disconnects background tabs.

### DIFF
--- a/src/device/webusb.ts
+++ b/src/device/webusb.ts
@@ -23,6 +23,7 @@ import {
   FlashDataSource,
   HexGenerationError,
   MicrobitWebUSBConnectionOptions,
+  SerialOption,
   WebUSBError,
 } from "./device";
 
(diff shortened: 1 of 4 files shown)


src/device/webusb.ts
```diff
@@ -78,7 +79,11 @@ export class MicrobitWebUSBConnection
         this.visibilityReconnect = false;
         if (!this.flashing) {
           this.log("Reconnecting visible tab");
-          this.connect();
+          this.connect({
+            // If any other connection status change occurs then visibilitReconnect is set to false, so
```
microbit-matt-hillsdon, line 83:
```suggestion
            // If any other connection status change occurs then visibilityReconnect is set to false, so
```

typo

src/serial/SerialArea.tsx
```diff
@@ -53,31 +53,31 @@ const SerialArea = ({
...
-              ml={1}
-              mr={1}
-              fontSizePt={terminalFontSizePt}
-              tabOutRef={tabOutRef}
-            />
-          </Box>
-        )}
+        <Box
+          // Need to render this when not connected as we need it to maintain scrollback across disconnect/reconnect in some cases.
+          display={connected ? undefined : "none"}
```
microbit-matt-hillsdon, line 58: Until this change reset wasn't really what was resetting the terminal, as we were also discarding it entirely.

Need to review for unexpected consequences here.

microbit-matt-hillsdon:
This will instead be fixed by #1238 once the connection library is released. Closing.

PR #1118, opened 2023-03-31
Deploy/invalidate REVIEW stage even if e2e fails

microbit-matt-hillsdon:
Often looking at that deployment is the quickest way to understand a failure.


microbit-robert (approved):
LGTM.

PR #1122, opened 2023-05-22
Update fragile WebUSB error matching code.

microbit-matt-hillsdon:
The text now appears to be prefixed, we have e.g.

> DOMException: Failed to execute 'requestDevice' on 'USB': No device selected.

Probably still worth it unfortunately.


microbit-matt-hillsdon:
> Once built I'll ensure we can trigger each case.

Confirmed.

microbit-robert (approved):
LGTM.

PR #1130, opened 2023-09-21
Upgrade to Node 20

microbit-matt-hillsdon:
Drop webusb dep that's only used for some test we largely didn't complete. Replace the working test with a mock as its use of the webusb dep wasn't meaningful.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Configure node
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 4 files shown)


microbit-matt-hillsdon:
@microbit-robert I've fixed the build issue.

microbit-matt-hillsdon:
Remaining `npm audit` issues are in react-scripts (so dev time only) and will be addressed by moving to Vite at some point.

microbit-robert (approved):
Changes look good. Build looks good.

PR #1135, opened 2023-10-18
Add a limited drafts flag

microbit-matt-hillsdon:
Good enough to view new Reference sections assuming they're published but the top-level document that adds them is a draft.

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -17,29 +17,33 @@ import { Stage, stage as stageFromEnvironment } from "./environment";
  */
 export type Flag =
   /**
-   * Enables verbose debug logging to the console of drag events.
-   */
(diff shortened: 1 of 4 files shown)


src/flags.ts
```diff
@@ -48,10 +50,11 @@ interface FlagMetadata {
 
 const allFlags: FlagMetadata[] = [
   // Alphabetical order.
```
microbit-matt-hillsdon, line 52: Yeah so not so much before 😄

microbit-matt-hillsdon:
You can now tell this is working as you can see the work-in-progress Neopixels section.

microbit-robert (approved):
LGTM.

PR #1141, opened 2023-12-12
Brand colour tweaks

microbit-matt-hillsdon:
I've aligned the tints/shades with the brand guidelines. There are some minor visual differences that we'll discuss internally.


microbit-matt-hillsdon:
Notable changes (all screenshots from new):

Screenshot2023_12_12_151345

better?

Screenshot2023_12_12_151406

hover state unintentionally broken before

image

worse? this was a purple/teal blend which makes some sense as it's displayed alongside the gradient
we'd need to just have this as a special case somehow on neither colour scale

microbit-robert:
I think the search highlighting is now too dark:
image

PR #1143, opened 2024-02-13
Build updates

microbit-matt-hillsdon:
Fixes issues encountered building on a new local install with Node 20


microbit-robert (approved):
LGTM.

PR #1144, opened 2024-02-20
Align menu items across apps

microbit-robert:
Bump theme version

Depends on https://github.com/microbit-foundation/python-editor-v3-microbit/pull/35


microbit-robert:
Will need to update the ui.en.json file in CrowdIn.

PR #1151, opened 2024-03-11
e2e fixes

microbit-matt-hillsdon:
- Catch another error case that can happen due to LS dispose
- Bump timeouts while we're look in into issues so we can be sure it's not just slow


src/e2e/app.ts
```diff
@@ -195,10 +195,10 @@ export class App {
...
       runBeforeUnload: true,
     });
-    // A delay is required to give us a chance to handle the dialog event.
-    // If this proves fragile we could wait for the success condition below
-    // and give up on testing the absence of a dialog.
-    await page.waitForTimeout(50);
+    await waitFor(() => {
+      expect(page.isClosed()).toEqual(true);
+    }, defaultWaitForOptions);
+    console.log(this.dialogs);
```
microbit-matt-hillsdon, line 201:
```suggestion
```

PR #1152, opened 2024-03-11
Vite

microbit-matt-hillsdon:


index.html
```diff
@@ -0,0 +1,55 @@
...
+    <script>
+      // GA config itself is via the common assets config depending on consent.
+      window.dataLayer = window.dataLayer || [];
+      window.gtag =
+        window.gtag ||
+        function () {
+          window.dataLayer.push(arguments);
+        };
+    </script>
+    <% } %>-->
```
microbit-matt-hillsdon, line 45: Need an approach to templating this

index.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="shortcut icon" href="/favicon.ico" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
+    <meta name="theme-color" content="#000000" />
```
microbit-matt-hillsdon, line 10: Is this right?

src/fs/migration.ts
```diff
@@ -36,7 +34,10 @@ export const parseMigrationFromUrl = (
   try {
     if (urlPart) {
       const bytes = toByteArray(urlPart);
-      const json = JSON.parse(LZMA.decompress(bytes));
+      const json = JSON.parse(
+        // FIXME, reinstate decompression with a working module.
```
microbit-matt-hillsdon, line 38: Needs an alternative lib or some way to import this old-style module.

vite.config.ts
```diff
@@ -0,0 +1,50 @@
...
+    testTimeout: 60_000,
+    hookTimeout: 30_000,
+  };
+  const config: UserConfig = {
+    base: process.env.BASE_URL ?? "/",
+    build: {
+      outDir: "build",
+    },
+    worker: {
+      format: "es",
```
microbit-matt-hillsdon, line 33: Is this an issue for browser support? Might break Safari 14 which otherwise has OK support: https://caniuse.com/mdn-api_worker_worker_ecmascript_modules

src/fs/migration.ts
microbit-matt-hillsdon: micro:bit classroom manages this but uses a different entry point (the worker one) and copes with different behaviour in dev vs build. Can we find an alternative?

microbit-matt-hillsdon:
Preview: https://review-python-editor-v3.microbit.org/vite/

index.html
microbit-matt-hillsdon: 7846276a03b11f1644e64229ca435e20aae8444a

vite.config.ts
microbit-matt-hillsdon: Confirmed this breaks search in Safari 14.1 as the worker fails to initialize.

microbit-matt-hillsdon: It works in Safari 15 even on iOS.

With type: "module" in the worker init Vite will still try to bundle to an iife by default.
However, that doesn't work if the worker itself dynamically loads code.
We do that for the stemmers as there's a blob of code/data per language. There's not an obvious way forward here.

One approach could be to disable search on platforms where the worker init fails. We already disable search in for languages that don't have Lunr support.

Perhaps it's possible to build a worker per language instead? That might be quite a change. Though I suppose it could be simple if verbose to have a worker file per language with the relevant import. Would need app changes to cope as previously the language change was internal to the worker.

microbit-matt-hillsdon: > Perhaps it's possible to build a worker per language instead

Did this. A bit messy but it should be quicker to init as we remove a waterfall step. 
Looks good on Safari 14.

.github/workflows/build.yml
```diff
@@ -36,20 +36,20 @@ jobs:
...
         if: github.repository_owner == 'microbit-foundation'
       - run: npm run ci
         env:
-          REACT_APP_GA_COOKIE_PREFIX: ${{ secrets.REACT_APP_GA_COOKIE_PREFIX }}
-          REACT_APP_GA_MEASUREMENT_ID: ${{ secrets.GA_MEASUREMENT_ID }}
-          REACT_APP_SENTRY_DSN: ${{ secrets.REACT_APP_SENTRY_DSN }}
-      - run: mkdir -p /tmp/app${PUBLIC_URL} && cp -r build/* /tmp/app${PUBLIC_URL} && npx serve --no-clipboard -l 3000 /tmp/app &
+          VITE_GA_COOKIE_PREFIX: ${{ secrets.VITE_GA_COOKIE_PREFIX }}
+          VITE_GA_MEASUREMENT_ID: ${{ secrets.GA_MEASUREMENT_ID }}
+          VITE_SENTRY_DSN: ${{ secrets.VITE_SENTRY_DSN }}
```
microbit-matt-hillsdon, line 51: Secrets will need updating

.gitignore
```diff
@@ -1,5 +1,7 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+.env
+
```
microbit-matt-hillsdon, line 4: Unusually we commit .env for this project so this ignore is wrong. I don't think we even build without it being persent.

I'll have a think about this as I'm not sure a mix of approaches across projects is a good idea.

index.html
```diff
@@ -2,35 +2,48 @@
...
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
     <meta name="theme-color" content="#6c4bc1" />
     <title>micro:bit Python Editor</title>
     <meta property="og:title" content="micro:bit Python Editor" />
+    <link rel="canonical" href="%VITE_FULL_URL%" />
+    <meta property="og:url" content="%VITE_FULL_URL%" />
+    <meta name="twitter:card" content="summary_large_image" />
+    <meta name="twitter:site" content="@microbit_edu" />
```
microbit-matt-hillsdon, line 16: Unintended change

```suggestion
```

index.html
```diff
@@ -2,35 +2,48 @@
...
-    />
-    <meta name="twitter:card" content="summary_large_image" />
     <meta
       name="twitter:description"
       content="A Python Editor for the BBC micro:bit, built by the Micro:bit Educational Foundation and the global Python Community."
     />
-    <% if (process.env.REACT_APP_FOUNDATION_BUILD === 'true') { %>
-    <script src="https://shared-assets.microbit.org/common/v1/common.js" async></script>
-    <link rel="stylesheet" href="https://shared-assets.microbit.org/common/v1/main.css"></link>
+    <meta property="og:image" content="%VITE_FULL_URL%social.png" />
```
microbit-matt-hillsdon, line 25: I've checked this on the review branch: https://review-python-editor-v3.microbit.org/vite/social.png

.github/workflows/build.yml
microbit-matt-hillsdon: GA ones are no longer in use.
Sentry updated.

.gitignore
microbit-matt-hillsdon: The Vite way seems to be to commit defaults in .env and use .env.local for uncommitted changes. So this is fine and I might update the approach elsewhere.

microbit-matt-hillsdon: The ignore however is wrong

microbit-matt-hillsdon, line 4:
```suggestion
```

src/fs/host.ts
```diff
@@ -87,18 +87,23 @@ export class IframeHost implements Host {
...
+  async shouldReinitializeProject(_storage: FSStorage): Promise<boolean> {
     // If there is persistence then it is the embedder's problem.
     return true;
   }
   createInitialProject(): Promise<PythonProject> {
     return new Promise((resolve) => {
-      this.window.addEventListener("load", () =>
-        notifyWorkspaceSync(this.parent)
-      );
+      // TODO: review this carefully!
```
microbit-grace, line 99: Does this still need reviewing carefully?

src/fs/migration.test.ts
```diff
@@ -4,13 +4,9 @@
...
 import { isMigration, parseMigrationFromUrl } from "./migration";
+import { testMigrationUrl } from "./migration-test-data";
 
-// The heart project.
-export const testMigrationUrl =
-  // origin needs to match jest's testUrl
-  "http://localhost/#import:#project:XQAAgACRAAAAAAAAAAA9iImmlGSt1R++5LD+ZJ36cRz46B+lhYtNRoWF0nijpaVyZlK7ACfSpeoQpgfk21st4ty06R4PEOM4sSAXBT95G3en+tghrYmE+YJp6EiYgzA9ThKkyShWq2UdvmCzqxoNfYc1wlmTqlNv/Piaz3WoSe3flvr/ItyLl0aolQlEpv4LA8A=";
-
-describe("parseMigrationFromUrl", () => {
+describe.skip("parseMigrationFromUrl", () => {
```
microbit-grace, line 9: Does this test still need skipping?
I removed the skip and ran it locally, and it passed :)

```suggestion
describe("parseMigrationFromUrl", () => {
```

microbit-grace (commented):
Mostly LGTM, left a few comments. Will leave for Rob to review

src/fs/host-default.test.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @vitest-environment-options { "url": "http://localhost:3000" }
+ */
+import { fromByteArray } from "base64-js";
+import { MAIN_FILE } from "./fs";
+import { DefaultHost } from "./host";
+import { defaultInitialProject } from "./initial-project";
+import { testMigrationUrl } from "./migration-test-data";
+
+describe("DefaultHost", () => {
+  it.skip("uses migration if available", async () => {
```
microbit-robert, line 12: Is this skipping on purpose or can it be reinstated?

src/fs/migration.test.ts
microbit-robert, line 9: Is this skipping on purpose or can it be reinstated?

src/language-server/pyright.ts
```diff
@@ -27,15 +31,21 @@ let cachedLang: string | undefined;
...
 ): Promise<LanguageServerClient | undefined> => {
-  // For jest.
+  // For jsdom.
   if (!window.Worker) {
     return undefined;
   }
-  if (cached && cachedLang === language) {
-    return cached;
+  if (cache) {
+    // This is safe to called if already initialized.
```
microbit-robert, line 39:
```suggestion
    // This is safe to call if already initialized.
```

src/fs/migration.test.ts
microbit-matt-hillsdon: Oops, no! This was disabled until I worked around the LZMA issues.

src/fs/host-default.test.ts
microbit-matt-hillsdon, line 12:
```suggestion
  it("uses migration if available", async () => {
```

microbit-matt-hillsdon: Nope, same issue as Grace's comment but different test. Fixed.

microbit-matt-hillsdon:
> I've seen a couple of these errors on switching language that I would guess we expect

I didn't expect these... will try to understand.

> We also get some shift in the sidebar with reference panel showing on changing language. Not a big issue, but a difference from production.

That's very odd. We do have some media-query-dependent styling in DocumentationTopLevelItem.

src/fs/host.ts
microbit-matt-hillsdon: This is OK (think it only happens in the unit tests anyway). Tried locally with classroom.
Will check classroom and Python integration again after merge.

microbit-matt-hillsdon: Comment removed in 9dc4ea0fe1556eee547d3ceba2e4f2a482a56186

microbit-matt-hillsdon:
> > We also get some shift in the sidebar with reference panel showing on changing language. Not a big issue, but a difference from production.
> 
> That's very odd. We do have some media-query-dependent styling in DocumentationTopLevelItem.

This isn't language-change specific. It happens on load too.

microbit-matt-hillsdon:
> > > We also get some shift in the sidebar with reference panel showing on changing language. Not a big issue, but a difference from production.
> > 
> > 
> > That's very odd. We do have some media-query-dependent styling in DocumentationTopLevelItem.
> 
> This isn't language-change specific. It happens on load too.

Fixed in d4313dd657a5f53609758e23b10da1e2e342b6f7. I think this was a Chakra UI upgrade issue rather than a Vite one.

microbit-matt-hillsdon:
I've so far failed to reproduce the error in @microbit-robert's screenshot. I do see this though, for example when starting in Dutch and switching to English (but not the other way around).

```
lunr.multi.js:50 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'wordCharacters')
    at t.multiLanguage (lunr.multi.js:50:58)
    at be (search.worker.ts:285:31)
    at Be.index (search.worker.ts:358:25)
    at Be.ctx.onmessage (search.worker.ts:344:21)
```

microbit-matt-hillsdon:
> I've so far failed to reproduce the error in @microbit-robert's screenshot. I do see this though, for example when starting in Dutch and switching to English (but not the other way around).
> 
> ```
> lunr.multi.js:50 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'wordCharacters')
>     at t.multiLanguage (lunr.multi.js:50:58)
>     at be (search.worker.ts:285:31)
>     at Be.index (search.worker.ts:358:25)
>     at Be.ctx.onmessage (search.worker.ts:344:21)
> ```

Fixed in https://github.com/microbit-foundation/python-editor-v3/pull/1152/commits/994c4a7dd918877734e1d04a61a405fb726a8bf0

src/documentation/search/search.worker.nl.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * (c) 2022, Micro:bit Educational Foundation and contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+import { SearchWorker } from "./search.worker";
+import languageSupport from "@microbit/lunr-languages/lunr.nl";
+if (!languageSupport) {
+  throw new Error("Whoops!");
```
microbit-robert, line 9: Whoops!

microbit-matt-hillsdon: 51504432bf84a757e19b0b59ba1d01ab3b5ae6c2

microbit-robert (approved):
LGTM.

PR #1153, opened 2024-03-15
Playwright

microbit-grace:


package.json
```diff
@@ -59,6 +60,7 @@
     "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
     "@chakra-ui/cli": "^2.4.1",
     "@formatjs/cli": "^6.2.7",
+    "@playwright/test": "1.33.0",
```
microbit-grace, line 63: The latest version of playwright should be `1.42.1`, but because the node version is v20.6.1, we get an error when running the tests (see [issue](https://github.com/microsoft/playwright/issues/29253)). So I have forced the v1.33.0. This version is high enough for the vite to be able to run as a webserver without [this issue](https://github.com/microsoft/playwright/issues/21227), but low enough to overcome the node issue

package.json
```diff
@@ -99,7 +101,7 @@
     "serve": "npx serve --no-clipboard -l 3000 -- build/",
     "start": "vite dev",
     "test:e2e:headless": "cross-env E2E_HEADLESS=1 vitest --mode e2e",
```
microbit-matt-hillsdon, line 103: Should we just remove this task? What did we do for data.microbit.org? Let's align on whatever you feel makes sense.

microbit-matt-hillsdon: I guess there must also be some vite/vitest config that we don't need anymore.

src/e2e/multiple-files.test.ts
```diff
@@ -3,60 +3,56 @@
...
   });
 
-  it("Shows warning for third-party module", async () => {
+  test("Shows warning for third-party module", async ({ app }) => {
     await app.loadFiles("testData/module.py", {
       acceptDialog: LoadDialogType.CONFIRM,
     });
-    await app.switchToEditing("module.py");
+    await app.editFile("module.py");
     await app.findThirdPartyModuleWarning("a", "1.0.0");
```
microbit-matt-hillsdon, line 51: This seems like the only thing still named findXXX. Fix for consistency?

microbit-matt-hillsdon (changes requested):
Please can you move playwright from package.json dependencies to dev dependencies and remove puppeteer and related packages?

Some other comments inline.

package.json
microbit-grace: Makes sense. About removing a task, is this what you mean 8c5684f6cb524b038bba44d3bdb4527a700001de? Or did you mean removing something else?

src/e2e/multiple-files.test.ts
microbit-grace: 179e918954ddcdffc85aaf2646eaf09c354e6dd3

src/e2e/app.ts
```diff
@@ -34,1429 +22,738 @@ export interface BrowserDownload {
...
   }
 
-  async awaitAnimation(): Promise<void> {
-    const page = await this.page;
-    await page.waitForTimeout(300);
+  async typeInEditor(text: string): Promise<void> {
+    const textWithoutLastChar = text.slice(0, text.length - 3);
+    await this.editorTextArea.fill(textWithoutLastChar);
+    // Last few characters are typed separately to trigger editor suggestions
+    const [a, b, c] = text.slice(-3);
```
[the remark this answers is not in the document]
microbit-grace: Have refactored `typeInEditor` here: https://github.com/microbit-foundation/python-editor-v3/pull/1153/commits/2339e62c825de3547e28fda53b9894786f886bd0

I changed it to slice the last 2 characters instead of last 3 since that seems to be sufficient for making the autocomplete tests pass

From observation, the delays don't feel significant. The delays were also added to reduce flakiness. I have updated the comment to mention this: https://github.com/microbit-foundation/python-editor-v3/pull/1153

package.json
microbit-matt-hillsdon: Yeah, either update it or remove it as makes sense to you. Should have said npm script not task. Resolving.

microbit-matt-hillsdon (approved):
Looks great, thanks.

PR #1156, opened 2024-03-19
[DO NOT MERGE] Preview of forthcoming beta MicroPython

microbit-matt-hillsdon:
We're doing this as a preview of the forthcoming beta. 
Stubs are a copy for now - we'll update them shortly. No sim changes yet.
MicroPython is not the official beta yet - that's not been released. We're using a hex from the tip of audio-recording.


src/simulator/Simulator.tsx
microbit-grace, line 1: File changes to undo once sim changes are in staging

PR #1157, opened 2024-03-25
Get the welcome video via the deployment rather than the CMS

microbit-matt-hillsdon:
This makes more sense than using the CMS as it's not really content in the same way as everything else.
Removes loading and error cases from the UI.

Motivated by https://github.com/microbit-foundation/python-editor-v3/issues/1148 as this isn't a bit of Sanity schema that makes any sense to extract.


microbit-robert (approved):
LGTM.

PR #1162, opened 2024-04-03
Screen height on tablet vs keyboard

microbit-grace:


src/common/VisualViewportCSSVariables.tsx
```diff
@@ -13,28 +13,31 @@ const isIOS = (): boolean =>
...
+  };
+
   useEffect(() => {
-    // If we remove the iOS check then we should look carefully at resize performance.
-    if (!window.visualViewport || !isIOS()) {
+    // svh on Safari in iOS does not resize when virtual keyboard is present
+    if (CSS.supports("height: 100svh") && !isIOS) {
+      addStyleInnerText(":root { --vvh: 100svh }");
+      return;
+    }
```
microbit-grace, line 34: Interestingly, `100svh` on Safari in iOS does not resize when virtual keyboard is present. I tried it in one of the commits. This is also somewhat also documented here: https://github.com/bramus/viewport-resize-behavior/blob/main/explainer.md#the-side-effect-of-a-resized-icb

src/common/VisualViewportCSSVariables.tsx
```diff
@@ -13,28 +13,31 @@ const isIOS = (): boolean =>
...
+    if (!window.visualViewport) {
       return;
     }
```
microbit-grace, line 37: Previously, we only used visualViewport when it is iOS. Was not sure why we limited to only iOS in the first place, since `window.visualViewport` appears to work well in Android as well 🤔 I used it as a fallback option for Android as well

src/common/SplitView/SplitViewSized.tsx
```diff
@@ -38,6 +38,7 @@ const SplitViewSized = ({ children }: SizedPaneProps) => {
   return (
     <Box
       pointerEvents={dragging ? "none" : "unset"}
+      display={mode === "collapsed" ? "none" : undefined}
       visibility={mode === "collapsed" ? "hidden" : undefined}
```
microbit-grace, line 42: This fixes some of the extra space around the UI that is added when the simulator is collapsed. `visibility: hidden` does not change the layout of the document, `display: none` does.

src/common/VisualViewportCSSVariables.tsx
microbit-grace, line 1: Not needed since we are aiming to have the keyboard overlay the UI instead of having a shifting action bar

PR #1164, opened 2024-04-08
Improve error handling for WebUSB issues

microbit-robert:
Show dialogs with updated and more user friendly text and graphics instead of a toast notification.
Continue to use existing translations until new text has been translated.

See https://microbit-global.monday.com/boards/1125389526/views/244984/pulses/1434828425 (private)


lang/ui.en.json
```diff
@@ -1116,29 +1116,21 @@
     "description": "Part of WebUSB error message"
   },
   "webusb-error-clear-connect-description-2": {
-    "defaultMessage": "Close any other tabs that may be using WebUSB (e.g. MakeCode, Python Editor), or unplug and replug the micro:bit before trying again.",
+    "defaultMessage": "Close any other tabs that may be using WebUSB (e.g. MakeCode, Python Editor, machine learning tool), or unplug and replug the micro:bit before trying again.",
```
microbit-matt-hillsdon, line 1119: I'm not sure that we want to add the ML tool here but one to confirm on Monday.

lang/ui.en.json
```diff
@@ -1116,29 +1116,21 @@
...
+    "defaultMessage": "Another process is connected to this micro:bit",
     "description": "Title of error for WebUsb connection"
   },
   "webusb-error-default-title": {
     "defaultMessage": "WebUSB error",
     "description": "Default title for error messages relating to WebUSB"
   },
   "webusb-error-reconnect-microbit-description": {
-    "defaultMessage": "Please reconnect your micro:bit and try again.",
+    "defaultMessage": "<p>Please:</p><ul><li>check that this micro:bit does not have a battery pack connected</li><li>unplug and replug the USB cable</li></ul>",
```
microbit-robert, line 1131: Do we want to update this text so it aligns more closely with what is shown in the video?

jaustin:
> ### Fallback error requesting reconnect
> image
> 
> * should this be a V2 micro:bit in the video?

Sure, I will update the video when I get a moment. Is it okay to swap the assets out at that point? How disruptive will it be if dimensions change slightly, etc?

I will also make sure '1' fades and '2' appears at the same time, as that was bothering me. Any other video feedback?

> 
> ### Failed to claim interface
> image
> 
> Should we change the language to not use process and maybe avoid the repetition? For example:
> 
> ```
> Another app or browser tab is connected to this micro:bit
> 
> Close any other tabs that may be using WebUSB (e.g. MakeCode, Python Editor, machine learning tool), or unplug and replug the micro:bit before trying again.
> ```

That language sounds better. To be even more helpful for the user, could we include favicon hints for each? 

image

image

image


> 
> Is app the best option? "application"? I think "program" isn't quite right especially in the context of a micro:bit editor and process is too techy.

I think conventionally it would be a 'page' right? Another  "Another page or browser tab is connected to this micro:bit" (I think even web apps are still web pages?)

> 
> Should this have the full replug advice + video? Leaning towards no as the best correction is likely to close the other tab.
> 

I agree no. I think it's common you need to find that other tab. Though I don't know when it's MakeCode or not. I *think* it's when you have multiple windows and MakeCode is the frontmost tab in one of them?

> ### Connection timeout
> image
> 
> I think this should have the replug advice + video.

Agree

microbit-robert:
Updated screenshots:

I have updated the text for this one so matches what the video shows. It's now quite tall and will likely scroll vertically on screen heights of 768px.
image

Improved title, simplified the description to lose reference to the machine learning tool.
image

Timeout error is now identifical to the reconnect error with a different title. Again, it's tall.
image

microbit-robert:
Alternative layouts

Horizontal:
image

Vertical with a lot of horizontal padding around the video:
image

lang/ui.en.json
microbit-robert: I've done this, as noted in the screenshots below.

lang/ui.en.json
microbit-robert: This is still unresolved, but I've simplified this back to being just MakeCode and the Python Editor for now.

PR #1166, opened 2024-04-10
feat: replace webusb with file system api

felixerdy:


felixerdy:
Sorry, this was meant to go against our fork.

PR #1168, opened 2024-04-15
Replace to use lunr-languages

microbit-grace:
instead of @microbit/lunr-languages forked package


microbit-matt-hillsdon (approved):
LGTM

PR #1169, opened 2024-04-16
In-context translation

microbit-grace:
Requires a translate flag.

Foundation builds only so we don't get confusing contributions to our Crowdin from forks that might have made text changes. Open to discussing making the Crowdin project configurable if useful to forks.

When the flag is present, change language id to `lol` and append CrowdIn JIPT script to header.

**Related PRs (now merged)**
Pyright - https://github.com/microbit-foundation/pyright/pull/32
Stubs - https://github.com/microbit-foundation/micropython-microbit-stubs/pull/100


src/workbench/Workbench.tsx
```diff
@@ -60,6 +61,16 @@ const Workbench = () => {
     }
   }, [selection, setSelectedFile, files]);
 
+  useEffect(() => {
+    if (flags.translate) {
```
microbit-matt-hillsdon, line 65: Is this adding it twice in development? I don't think we can meaningfully clean it up (it'll still have been loaded if we remove the element) but we could add an id to it and check whether it already existed.

microbit-grace: a753d1cd30908ddb693f8904adcc9a6449ed1270

microbit-matt-hillsdon (approved):

PR #1170, opened 2024-04-17
PWA functionality including offline mode (featured flagged)

microbit-robert:
Notable behaviours:
- Fallback to English when changing language offline (when the selected language assets aren’t cached).
- Shows image placeholders for uncached images when offline.
- The app will automatically trigger a page reload when it detects the service worker has been updated. The service worker is updated when any of the build assets are updated.

There are two levels of feature gating:
- It's only enabled for microbit-foundation builds to avoid others shipping it accidentally. You can change FEATURE_PWA in the build to change this. We're open to a better system of build-time feature flags if there's interest (at least allowing existing env vars to override would make sense).
- There's a client side feature flag. You have to set this via `localStorage.setItem("flags", "pwa")` for now if you want to test an actual PWA install. We'll default that flag on to release to the beta after building more confidence internally.


microbit-matt-hillsdon:
> all available languages are precached

Does that mean everyone pays the cost of every language now? Before interactivity or after/in the background?

vite.config.ts
```diff
@@ -52,6 +53,92 @@ export default defineConfig(({ mode }) => {
...
+        registerType: "autoUpdate",
+        devOptions: {
+          enabled: true,
+        },
+        workbox: {
+          maximumFileSizeToCacheInBytes: 3097152,
+          globPatterns: ["**/*.{js,css,html,ico,png,svg,gif,hex}"],
+          runtimeCaching: [
+            {
+              urlPattern: /^https:\/\/ajwvhvgo.apicdn.sanity.io\/.*/,
```
microbit-matt-hillsdon, line 66: Only worth addressing after other issues: we've parameterised the project name and dataset elsewhere so should do that here too (and below).

microbit-robert:
> > all available languages are precached
> 
> Does that mean everyone pays the cost of every language now? Before interactivity or after/in the background?

For this version, yes. We need to be more specific in what we precache. Maybe nothing language specific and save that for runtime. It does feel like the plugin might not suit our needs. We'll either end up creating just the service worker from scratch, or ditch the plugin and do everything from scratch.

src/simulator/Simulator.tsx
```diff
@@ -43,14 +43,15 @@ const Simulator = ({
...
+  //   "https://python-simulator.usermbit.org/v/0.1/simulator.html";
+  // const staging =
+  //   "https://python-simulator.usermbit.org/staging/simulator.html";
+  // const url = stage === "PRODUCTION" ? production : staging;
   // For testing with sim branches:
-  //const branch = "upgrade-mpy";
-  //const url = `https://review-python-simulator.usermbit.org/${branch}/simulator.html`;
+  // TODO: Revert.
+  const branch = "service-worker";
+  const url = `https://review-python-simulator.usermbit.org/${branch}/simulator.html`;
```
microbit-robert, line 55: TODO: Revert before merge.

src/index.tsx
```diff
@@ -7,6 +7,37 @@ import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 import App from "./App";
 import reportWebVitals from "./reportWebVitals";
+import { registerSW } from "virtual:pwa-register";
+
+registerSW({
+  immediate: true,
+  // Cache runtime resources on first load.
+  // See https://github.com/GoogleChromeLabs/pwa-wp/issues/180.
+  onRegisteredSW(_, registration) {
```
microbit-robert, line 16: The best or most efficient way? It might be the only way.

microbit-robert:
Toast shown when changing language while offline and language assets are not already cached. Text is a working placeholder and needs discussion.

image

.github/workflows/build.yml
```diff
@@ -24,6 +24,8 @@ jobs:
       STAGING_CLOUDFRONT_DISTRIBUTION_ID: E2ELTBTA2OFPY2
       REVIEW_CLOUDFRONT_DISTRIBUTION_ID: E3267W09ZJHQG9
       VITE_FOUNDATION_BUILD: ${{ github.repository_owner == 'microbit-foundation' }}
+      # Optionally disable the PWA config (service worker and web manifest)
+      # DISABLE_PWA: '1'
```
microbit-robert, line 28: Let's confirm what the default behaviour should be.

src/flags.ts
```diff
@@ -60,6 +67,7 @@ const allFlags: FlagMetadata[] = [
   { name: "noLang", defaultOnStages: [] },
   { name: "translate", defaultOnStages: [] },
   { name: "noWelcome", defaultOnStages: ["local", "REVIEW"] },
+  { name: "pwa", defaultOnStages: ["REVIEW"] },
```
microbit-robert, line 70: This needs more thought/discussion. If you install the PWA, close it, then reopen it, you'll lose the flag and the ability to add the flag. In the meantime, this change will make the review branch a bit more useful for testing.

vite.config.ts
```diff
@@ -28,7 +29,29 @@ const viteEjsPlugin = ({ data }: { data: ejs.Data }): Plugin => ({
...
+    handler: (
+      html: string,
+      _ctx: IndexHtmlTransformContext
+    ): IndexHtmlTransformResult => {
+      return html.replace(
+        `<link rel="manifest" href="${
+          process.env.BASE_URL ?? "/"
+        }manifest.webmanifest">`,
+        ""
+      );
```
microbit-robert, line 48: Hmmm. This works, but feels fragile.

src/language-server/client.ts
```diff
@@ -178,7 +185,16 @@ export class LanguageServerClient extends EventEmitter {
...
         }
-        throw e;
+        if (!navigator.onLine) {
+          showOfflineLanguageToast(this.toast);
+          // Fallback to the precached locale if user is offline.
+          // Currently unused as we are precaching all pyright-locale files.
+          this.locale = fallbackLocale;
+          this.initializePromise = undefined;
+          this.initialize();
+        } else {
```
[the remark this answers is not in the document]
microbit-matt-hillsdon: See https://github.com/microbit-foundation/pyright/pull/34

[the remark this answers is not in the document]
microbit-robert: We now only precache the fallback language for pyright using the build of the PR above [d45eeae](https://github.com/microbit-foundation/python-editor-v3/pull/1170/commits/d45eeae52e485ca75433c825fdb5e199402aa13d)

microbit-matt-hillsdon:
image

This feels a bit odd when in PWA mode as there's no obvious affordance to reload the page. Ideally we'd find a more neutral phrasing.

src/index.tsx
```diff
@@ -7,6 +7,49 @@ import { StrictMode } from "react";
...
+                  payload: { urlsToCache },
+                });
+              }
+            };
+          }
+        };
+      }
+    },
+  });
+}
```
microbit-matt-hillsdon, line 52: Do we need an `else` here similar to the sim?
Scenario to test: use flag, update app, then don't use flag and should see latest

vite.config.ts
microbit-matt-hillsdon: I think we should assert that this changes html.

vite.config.ts
```diff
@@ -52,6 +75,98 @@ export default defineConfig(({ mode }) => {
       }),
       react(),
       svgr(),
+      VitePWA({
```
microbit-matt-hillsdon, line 78: Could baseUrl or a similar option partition the service worker caches for beta/v3 ?

microbit-robert: Maybe workbox.cacheId will work (main/beta or prod/staging)
https://developer.chrome.com/docs/workbox/modules/workbox-build#type-GeneratePartial

microbit-matt-hillsdon: Trying that now...

.github/workflows/build.yml
microbit-matt-hillsdon: I've flipped it to enable only for Foundation builds so that folks (especially students) don't get bitten by the SW when doing more experimental work.

vite.config.ts
microbit-matt-hillsdon: This looks good. We had to explicitly name the non-pre-cache caches.

src/index.tsx
microbit-matt-hillsdon: 61213d240c8931b5636bbfb739fcd7f3fa74cda8

microbit-matt-hillsdon (approved):

microbit-matt-hillsdon:
I'll merge this tomorrow when I'm more available for any fallout.

microbit-matt-hillsdon:
This isn't working in practice as we used scoped service workers but the scope is /v/beta/ but typically access is at /v/beta for staging/production. Ideally we wouldn't redirect to the less pretty URLs.

We could register against a slashless scope by adding a header. It's a prefix match so that's a bit awkward, e.g. /v/3 would also match /v/31 and /v/beta would match /v/betamax. The former seems worth some consideration. Alternatively we just have one and accept the extra reload when switching between versions.

microbit-matt-hillsdon:
> We could register against a slashless scope by adding a header. It's a prefix match so that's a bit awkward

We did this and it now seems to be working. Still behind the flag you need to set in localStorage.

PR #1174, opened 2024-04-22
[DRAFT / WIP - DO NOT MERGE] Show filtered builtins documentation

microbit-robert:
Also allows links to the builtins from autocomplete and signature help where documentation is available.

Using type stubs from https://github.com/microbit-foundation/micropython-microbit-stubs/pull/101


microbit-robert:
Unfortunate case where we are filtering out `__new__` (with other special methods), but they are the only methods documented in the class:
image

PR #1175, opened 2024-04-23
Centralize keyboard shortcuts

microbit-robert:


src/project/project-actions.tsx
```diff
@@ -76,6 +76,8 @@ import ProjectNameQuestion from "./ProjectNameQuestion";
  */
 export type LoadType = "drop-load" | "file-upload";
 
+export type FinalFocusRef = React.RefObject<HTMLElement>;
```
microbit-robert, line 79: We can now update this in one place in the future, yay!

src/common/keyboard-shortcuts.ts
```diff
@@ -0,0 +1,14 @@
+// Shortcuts are global unless noted otherwise.
+export const keyboardShortcuts = {
+  // This is scoped by keyboard focus.
+  copyCode: ["ctrl+c", "meta+c", "enter"],
+  search: ["ctrl+shift+f", "meta+shift+f"],
+  sendToMicrobit: ["ctrl+shift+e", "meta+shift+e"],
+  saveProject: ["ctrl+shift+s", "meta+shift+s"],
```
microbit-robert, line 7: Keybindings to discuss. I assumed we wanted a shortcut for save project rather than save file, but I suppose we could have both if we wanted.

microbit-robert:
Keyboard shortcut handlers are currently registered in the same components that host the UI for triggering the same action.

The Xterm keyboard shortcuts are already scoped to the serial terminals and it looks like more effort than it's worth to use this new hook for a custom copy handler and escape hatch.

src/documentation/common/CodeEmbed.tsx
```diff
@@ -180,7 +172,7 @@ const CodeEmbed = ({
           _focusVisible={{
             outline: "none",
           }}
-          onKeyDown={handleKeyDown}
+          // onKeyDown={handleKeyDown}
```
microbit-matt-hillsdon, line 175: What's intended here?

microbit-robert: A failure to fully remove the previous way of registering the relevant keydown handler.

microbit-robert: Removed here [e313a04](https://github.com/microbit-foundation/python-editor-v3/pull/1175/commits/e313a0484bb360d53342cf05b58c8de00cbe65c2).

src/common/keyboard-shortcuts.ts
microbit-matt-hillsdon: Yep we really want to encourage the hex to be what gets saved so that makes sense to me.

microbit-matt-hillsdon (approved):
LGTM

PR #1176, opened 2024-05-01
Upgrade CodeMirror

microbit-robert:


microbit-matt-hillsdon:
Did some testing on iPad and Mac Safari, Mac Firefox and Mac Chrome and all seems good.

microbit-matt-hillsdon:
I think we should get this out on beta and see what happens to logged errors there.

microbit-matt-hillsdon (approved):

PR #1177, opened 2024-05-02
Warn users when using V2 API features with micro:bit V1

microbit-robert:
Updates Pyright to included the changes from https://github.com/microbit-foundation/pyright/pull/35/


src/messages/ui.en.json
```diff
@@ -2430,7 +2448,7 @@
   "webusb-error-clear-connect-description-2": [
     {
       "type": 0,
-      "value": "Close any other pages or tabs that may be using WebUSB (e.g. MakeCode, Python Editor), or unplug and replug the micro:bit before trying again."
+      "value": "Close any other tabs that may be using WebUSB (e.g. MakeCode, Python Editor), or unplug and replug the micro:bit before trying again."
```
microbit-matt-hillsdon, line 2451: Puzzled about this. When did this change and why wasn't this file updated then?

microbit-matt-hillsdon: This change happens on main too.
I think it was accidentally included in 9bfff59d7ba02bce5c3f557c2b184c18ae1c99bb because we made a change to ui.en.json then reverted it but failed to revert the updated src/messages file. So this revert is fine.

microbit-matt-hillsdon (approved):
LGTM

PR #1180, opened 2024-05-23
[DRAFT - DO NOT MERGE] Use ml tool output with Python Editor POC

microbit-robert:
This is an alternative to https://github.com/PersonaOASIS/python-editor-v3/tree/ml-area that we can iterate on.

It uses the stubs from https://github.com/PersonaOASIS/micropython-microbit-stubs and the hex from https://github.com/PersonaOASIS/micropython-microbit-v2.

No consideration has been given to supporting other extensions yet.

`ml` has been renamed to `model` to match the stubs so that the links to API from the editor work.


src/project/project-actions.tsx
```diff
@@ -245,6 +248,35 @@ export class ProjectActions {
...
+    // Avoid lingering messages related to the previous project.
+    // Also makes e2e testing easier.
+    this.actionFeedback.closeAll();
+
+    if (files.length === 1 && getLowercaseFileExtension(files[0].name)) {
+      const file = files[0];
+      const json = await readFileAsText(file);
+      const data = JSON.parse(json) as ActionData[];
+      this.setMlData(data);
+      const actionNames = data.map((action) => action.name).join(" ");
```
microbit-robert, line 268: Action names can have spaces in so this will break. Probably better to have a name per line, but this will require a MicroPython update to change.

src/project/project-actions.tsx
```diff
@@ -259,7 +291,7 @@ export class ProjectActions {
    * @param files the files from drag and drop or an input element.
    * @param the type of user event that triggered the load.
    */
-  load = async (
+  loadHex = async (
```
microbit-robert, line 294: Not the best naming, as this function can handle more than just hex files, though it is its main purpose.

src/documentation/ModelArea.tsx
```diff
@@ -0,0 +1,223 @@
...
+      )}
+    </HeadedScrollablePanel>
+  );
+};
+
+interface MlItemProps {
+  content: ModelTopicEntry;
+}
+
+const ModelTopicEntry = ({ content }: MlItemProps) => {
```
microbit-robert, line 127: This whole component is hacky. If we had a markdown alternative for rendering extension documentation, we'd use it here. It's pretty horrible to write the model topic entries in the Sanity format.

PR #1181, opened 2024-05-28
Enable the simulator service worker on beta/staging

microbit-matt-hillsdon:
This is experimental at this point and is part of investigating PWA support for the micro:bit Python Editor.


microbit-robert (approved):
LGTM.

PR #1188, opened 2024-07-08
Use EventTarget rather than the Node "events" module

microbit-matt-hillsdon:
This should amount no meaningful change but is a useful preliminary to extracting the device code for reuse.

Incorporates the only meaningful file from https://github.com/DerZade/typescript-event-target/ (with licence details) to save on the dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "crelt": "^1.0.5",
     "dapjs": "2.2.0",
     "dompurify": "^2.3.3",
-    "events": "^3.3.0",
     "file-saver": "^2.0.5",
     "framer-motion": "^10.2.4",
     "lodash.debounce": "^4.0.8",
--- a/src/device/mock.ts
+++ b/src/device/mock.ts
@@ -3,18 +3,19 @@
  *
  * SPDX-License-Identifier: MIT
  */
+import { TypedEventTarget } from "../common/events";
 import {
   BoardVersion,
   ConnectionStatus,
   DeviceConnection,
-  EVENT_FLASH,
-  EVENT_SERIAL_DATA,
-  EVENT_STATUS,
+  DeviceConnectionEventMap,
   FlashDataSource,
+  FlashEvent,
+  SerialDataEvent,
+  StatusEvent,
   WebUSBError,
   WebUSBErrorCode,
 } from "./device";
-import EventEmitter from "events";
 
 /**
  * A mock device used during end-to-end testing.
@@ -24,7 +25,7 @@ import EventEmitter from "events";
  * the connected state without a real device.
  */
 export class MockDeviceConnection
-  extends EventEmitter
+  extends TypedEventTarget<DeviceConnectionEventMap>
   implements DeviceConnection
 {
   status: ConnectionStatus = navigator.usb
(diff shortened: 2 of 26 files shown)


src/device/webusb.ts
```diff
@@ -402,11 +402,11 @@ export class MicrobitWebUSBConnection
     if (this.device) {
       return this.device;
     }
-    this.emit(EVENT_START_USB_SELECT);
+    this.dispatchTypedEvent("start_usb_select", new StartUSBSelect());
     this.device = await navigator.usb.requestDevice({
       filters: [{ vendorId: 0x0d28, productId: 0x0204 }],
     });
-    this.emit(EVENT_END_USB_SELECT);
+    this.dispatchTypedEvent("start_usb_select", new EndUSBSelect());
```
microbit-matt-hillsdon, line 409: Wrong event name.

src/e2e/simulator.test.ts
```diff
@@ -5,16 +5,23 @@
  */
 import { test } from "./app-test-fixtures.js";
 
-const basicTest = "from microbit import *\ndisplay.show(Image.NO)";
+const basicTest = `from microbit import *
+display.show(Image.NO)`;
 
-const buttonTest =
-  "from microbit import *\nwhile True:\nif button_a.was_pressed():\ndisplay.show(Image.NO)";
+const buttonTest = `from microbit import *
```
microbit-matt-hillsdon, line 11: I've corrected the indentation in these tests which surprised me when I saw them in failure screenshots (for reasons unrelated to the indentation).

src/workbench/Workbench.tsx
```diff
@@ -23,43 +23,47 @@ import { SizedMode } from "../common/SplitView/SplitView";
...
+  const [maybeInvalidSelection, setSelection] = useSelection();
   const { files } = useProject();
+  const selection = defaultSelection(maybeInvalidSelection, files);
   const setSelectedFile = useCallback(
     (file: string) => {
       setSelection({ file, location: { line: undefined } });
     },
     [setSelection]
   );
-  useEffect(() => {
```
microbit-matt-hillsdon, line 53: I think this was a listener ordering thing. I've changed it to avoid making a state update at all.

src/device/mock.ts
```diff
@@ -49,9 +50,7 @@ export class MockDeviceConnection
 
   async initialize(): Promise<void> {}
 
-  dispose() {
-    this.removeAllListeners();
```
microbit-matt-hillsdon, line 53: We can't do this anymore as EventTarget doesn't have such API.

All listener add/removes are correctly paired except that in host.ts and a test, neither of which need to remove their listeners.

package.json
```diff
@@ -34,7 +34,6 @@
     "crelt": "^1.0.5",
     "dapjs": "2.2.0",
     "dompurify": "^2.3.3",
-    "events": "^3.3.0",
```
microbit-robert, line 37: Requires `npm i` and another commit to udpate `package-lock.json`.

microbit-matt-hillsdon: Oops, thanks

PR #1196, opened 2024-11-25
Queue problematic view dispatch call

microbit-robert:


microbit-matt-hillsdon:
@microbit-robert alternative approach on https://github.com/microbit-foundation/python-editor-v3/pull/1198. Still uses the same fix for this case but I think it removes to do anything unusual in other scenarios.

microbit-matt-hillsdon:
Closed as we went with https://github.com/microbit-foundation/python-editor-v3/pull/1198

PR #1197, opened 2024-11-27
Reinstate simulator compass animation

microbit-robert:
This must have regressed when we updated Chakra?

The following warning is logged when using the `as` prop, and is fixed by pass the icon as a child instead.
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()


microbit-matt-hillsdon:
Well spotted!

PR #1200, opened 2024-12-16
Fix missing "Code copied..." string

microbit-matt-hillsdon:
Closes https://github.com/microbit-foundation/python-editor-v3/issues/1199


microbit-robert (approved):
LGTM.

PR #1211, opened 2025-04-01
Update microbit-connection library to v0.0.0-alpha.36

microbit-grace:
Update connection library to include fix from https://github.com/microbit-foundation/microbit-connection/pull/55


microbit-robert (approved):
LGTM.

PR #1221, opened 2025-06-19
Upgrade to vite, vitest and vite plugins

microbit-grace:
[Task (Private)](https://microbit-global.monday.com/boards/1125389526/pulses/1906242065)

- [x] Test PWA and that it still works after changes


microbit-grace:
Have tried opening the PWA via preview link after making the changes and it seems to work as expected

microbit-matt-hillsdon:
Checked service worker upgrade on beta and all went well.

PR #1222, opened 2025-06-20
Fix escape crowdin jipt url

microbit-grace:
Redirect to a valid URL when deciding not to log into crowdin.

**Issue:** 

1. Visit https://python.microbit.org/v/3?flag=translate
2. Click on the close button instead of logging into Crowdin.
3. Get redirected to `https://python.microbit.org/v/%VITE_FULL_URL`, which is a request error


microbit-robert (approved):
LGTM

PR #1223, opened 2025-06-26
Remove main file if empty during hex creation

microbit-robert:
This creates a "clean" MicroPython hex. When the REPL is launched it already includes "from microbit import *".

### DIFF
--- a/src/fs/fs.ts
+++ b/src/fs/fs.ts
@@ -406,7 +406,10 @@ export class FileSystem extends TypedEventTarget<EventMap> {
 
   async statistics(): Promise<Statistics> {
     const fs = await this.initialize();
-    const currentMainFile = fs.readBytes(MAIN_FILE);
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
Please can we double check that we don't read main.py from the fs directly anywhere else?

microbit-robert:
> Please can we double check that we don't read main.py from the fs directly anywhere else?

We read it using FileSystem read, but not using MicropythonFsHex read/readBytes. So this looks safe to me.

microbit-robert:
The V2 editor also fails to load hex files without python code:
Screenshot 2025-06-26 at 14 07 51

microbit-robert:
This change also allows the user to create an additional file with content. If `main.py` is empty, the saved hex when re-imported will remove `main.py` from the project in the editor.

PR #1225, opened 2025-07-02
Fix in-context translation when no local storage settings

microbit-grace:
[Task (private)](https://microbit-global.monday.com/boards/1125389526/pulses/2029300298)

Fix settings default so that override is applied so that in-context translation works when no local storage setting is set.


microbit-matt-hillsdon (approved):
LGTM

PR #1238, opened 2026-03-06
Upgrade connection library to v1.0.0-beta.0 for serial fixes

microbit-matt-hillsdon:
This contains the pre-1.0 breaking API changes and USB serial re-work.

This should improve serial issues after first physical connection + flash and extra serial from previous programs leaking into subsequent ones.

I've also changed the default logger to a console logger as it's annoying to get less log output without the micro:bit theme - particularly important for testing the connection library and getting its logs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "@codemirror/view": "^6.26.3",
         "@emotion/react": "^11.11.4",
         "@emotion/styled": "^11.11.5",
-        "@microbit/microbit-connection": "^0.0.0-alpha.36",
+        "@microbit/microbit-connection": "^0.9.0-apps.alpha.17",
         "@microbit/microbit-fs": "^0.10.0",
         "@sanity/block-content-to-react": "^3.0.0",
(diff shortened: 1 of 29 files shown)


microbit-grace:
Mostly looks good. Found 2 things whilst testing: 

1. Switching tabs doesn't seem to pause serial in the same way we do in live. 
**Update [not an issue]**: I think it's because I had devtools opened with "Emulate a focused page" checked. Once I closed devtools or unchecked the option, it was working fine.

2. When I CTRL+D (soft reboot) in the serial window multiple times and then interrupt the program (CTRL+C), it soft reboots automatically after interruption. This happens in live. Not sure if it is a bug.

microbit-grace (approved):

microbit-grace:
Connect and flash error handling seems to match changes in DeviceErrorCode - https://github.com/microbit-foundation/python-editor-v3/blob/51f7678db087c9fe3c5b482911e46d587ec4bf6f/src/project/project-actions.tsx#L836-L872

PR #1240, opened 2026-04-30
Don't check board version if device is not connected

microbit-grace:
Since upgrading the connection library, getBoardVersion throws "Not connected" error if micro:bit is not connected. In this case, we only want to check the board version to warn users of v2-only features when a micro:bit is connected.

To repro issue, open https://python.microbit.org/v/beta, then drag and drop v2-only code snippet, such as a data logging code snippet. See error thrown in devtools console.

### DIFF
--- a/src/editor/codemirror/language-server/diagnostics.ts
+++ b/src/editor/codemirror/language-server/diagnostics.ts
@@ -8,6 +8,7 @@ import * as LSP from "vscode-languageserver-protocol";
 import { Action, Diagnostic } from "../lint/lint";
 import { positionToOffset } from "./positions";
 import { MicrobitUSBConnection } from "@microbit/microbit-connection/usb";
+import { ConnectionStatus } from "@microbit/microbit-connection";
 
 const reportMicrobitVersionApiUnsupported =
   "reportMicrobitVersionApiUnsupported";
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon (commented):
OK so one interesting aspect, we do cache after connection, at which point arguably we know which micro:bit you have even if you subsequently disconnect it. So perhaps this code would be better off error handling, to benefit from the caching? WDTY? Maybe:

```javascript
  let boardVersion: BoardVersion | undefined;
  try {
    boardVersion = device.getBoardVersion();
  } catch {
    // No device has ever been connected (or post clearDevice).
  }
  if (
    code === reportMicrobitVersionApiUnsupported &&
    (!warnOnV2OnlyFeatures || boardVersion !== "V1")
  ) {
    return undefined;
  }
```

Claude agrees there are no other unguarded connection calls that throw.

microbit-grace:
Had a play and I think that the UX is slightly better. It means that the V2-only warnings don't disappear suddenly when the user unplugs/disconnect their V1 micro:bit. If the user do find the warnings annoying because they have mistakenly plugged in their V1 and are intending to use the program for V2, they can "Disable warnings about V2-only features". 

However, this change in behaviour might mean changing the copy in the settings dialog:

Screenshot 2026-04-30 at 15 48 51

New copy along the lines of "Warnings are shown in the editor when a micro:bit V1 has been last connected"

microbit-matt-hillsdon:
Ah, I hadn't really twigged that the new approach matches the old behaviour, I was worried we'd made it worse in this respect. I don't think this is really worth changing copy for. Merging as is.

PR #1244, opened 2026-06-26
Lock background scroll while modals are open

microbit-matt-hillsdon:
Chakra 2.9 flipped the Modal preserveScrollBarGap default to true (chakra-ui/chakra-ui#8222), which maps to react-remove-scroll's removeScrollBar=false. That drops the body overflow:hidden lock, leaving the page scrollbar live and draggable behind the modal overlay. Set preserveScrollBarGap={false} on every Modal/Drawer/AlertDialog to restore the pre-2.10 hard scroll lock.


microbit-grace (approved):
LGTM

PR #1247, opened 2026-07-14
Secure npm dependencies and update packages

microbit-grace:
### Security hardening

- New .npmrc enforcing:
  - min-release-age=7 — a 7-day cooldown before a newly-published version can be installed (supply-chain protection), excluding our own @microbit/* and @microbit-foundation/* scopes
  - allow-git/remote/file/directory=root — blocks transitive dependencies from non-registry sources
  - strict-allow-scripts=true — every dependency lifecycle script must be explicitly allow-listed
  - engine-strict=true — enforces the engines versions
- allowScripts allowlist in package.json for the packages in our tree that request install scripts:
  - esbuild, fsevents → true (genuine native/build steps)
  - protobufjs, aws-sdk, core-js → false (cosmetic postinstalls — regenerating bundled files, maintenance/funding notices; the latter two come from the CI-only deploy tooling)
- engines bumped to node >=24.0.0, npm >=11.18.0 (required for allowScripts).

### Dependency update + type fixes

npm update bumped several packages. Fixes:

- lunr / @types/lunr — the global declare namespace lunr augmentation no longer merges into the imported binding, and @types/lunr's export = can't be module-augmented. Describe the multiLanguage + per-language plugins that lunr-languages attaches at runtime via a LunrWithLanguages type and cast at the call sites.
- @codemirror/view — ViewPlugin.extension now resolves to null unless the Arg type param is undefined; pin ViewPlugin.fromClass<…> so the drag-and-drop decorations plugin is an Extension again.
- vscode-languageserver-protocol — Diagnostic.message is now string | MarkupContent (extract the string); SignatureHelp.activeParameter can be null (widen guard to != null).
- vscode-jsonrpc — bumped the direct dependency to ^9 to dedupe against vscode-languageserver-protocol@3.18 (which requires vscode-jsonrpc@9), fixing the duplicate-MessageConnection type mismatch in pyright.ts.
- vitest 4 — the async rejects matcher dropped toThrowError; use toThrow.
- Fix hover over the dot in the margin no longer shows the diagnostic tooltip. @codemirror/view 6.43 hides any tooltip whose measured position falls outside the visible content area, unless Tooltip.clip === false
- Fix code mirror tooltip text spacing. Override whiteSpace: pre-line in code mirror tooltip

### Test config

- vscode-jsonrpc@9's ./browser subpath only declares a browser export condition. The app build applies it, but vitest resolves in node mode and can't — added a test-scoped alias in vite.config.ts pointing at the browser entry.

### CI (build.yml)

- Install npm@11 after setup-node so CI meets the new engines/allowScripts requirements.
- Bump the Playwright Docker image to v1.61.1-noble to match the Playwright version npm update resolved to (fixes "browser executable doesn't exist").

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,7 @@ jobs:
           node-version: 24.x
           cache: "npm"
           registry-url: "https://npm.pkg.github.com"
+      - run: npm install -g npm@11 --registry=https://registry.npmjs.org
       - run: npm ci
         env:
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
         if: env.STAGE == 'REVIEW' || env.STAGE == 'STAGING'
       - name: Run Playwright tests
         if: env.STAGE == 'REVIEW' || env.STAGE == 'STAGING'
-        uses: docker://mcr.microsoft.com/playwright:v1.58.2-noble
+        uses: docker://mcr.microsoft.com/playwright:v1.61.1-noble
         with:
           args: npx playwright test
       - name: Store reports
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,13 @@
+min-release-age=7
+min-release-age-exclude[]=@microbit/*
+min-release-age-exclude[]=@microbit-foundation/*
+
+# root here means this project's package.json
+allow-git=root
+allow-remote=root
+allow-file=root
+allow-directory=root
+
+strict-allow-scripts=true
+
+engine-strict=true
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,7 @@
         "react-icons": "^4.12.0",
         "react-intl": "^6.6.8",
         "vite": "^7.3.1",
-        "vscode-jsonrpc": "^8.0.0",
+        "vscode-jsonrpc": "^9.0.0",
         "vscode-languageserver-protocol": "^3.16.0",
         "web-vitals": "^1.1.1",
         "xterm": "4.14.1",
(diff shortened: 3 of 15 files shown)


microbit-matt-hillsdon:
The .npmrc parts of this all look good.

The lock file changes upgrade CodeMirror through quite a few minor versions at once. I think we need to do a manual testing pass for this to build confidence. Suggested areas:

- autocomplete
- signature help
- error diagnostics
- structure highlighting (including scrolling/resizing/editing)
- dnd of snippets
- and for the Lunr bits, check non-English search

microbit-matt-hillsdon:
Branch: https://review-python-editor-v3.microbit.org/npm-update/

Not sure why no PR link.

microbit-matt-hillsdon:
Old:
image

New:
image

Help/API are there if you scroll.

microbit-matt-hillsdon:
Hover over the dot in the margin no longer shows the diagnostic tooltip:

image

microbit-grace:
**Manual test checklist**

1. Autocomplete — @codemirror/autocomplete + language-server/autocompletion.ts
- [x] Type display. → completion li type, Enter/click inserts.
- [x] Completion docs panel renders alongside the list.
- [x] Trigger mid-identifier (ctrl + space) correctly (no clipping/offset when text is at the edge of the editor).

2. Signature help — language-server/d the activeParameter null guard)
- [x] Type display.scroll( → signature popup shows, active param is highlighted, and updates as you move across commas.
- [x] A function with no parameters still pops without error (was changed from undefined -> null case).

3. Error diagnostics — language-server/diagnostics.ts (message extraction changed  from string | MarkupContent)
- [x] Introduce a syntax/name error → red underline appears at the right span; hover shows the message text
- [x] V2-only API on a V1 board still shows the version warning; fix the error → underline clears.

4. Structure highlighting — structure-highlighting/ (viewport-driven ViewPlugin)
- [x] Nested blocks get their highlight bands. Then stress the viewport recomputation:
  - [x] Scroll a long file — bands ss.
  - [x] Resize the editor pane.
  - [x] Edit (add/remove indentation, split a block) — bands update live.

5. Drag-and-drop snippets — dnd.ts + dnd-decorations.ts (pinned the ViewPlugin.fromClass type arg)
- [x] Drag a snippet from the reference  —  preview highlight while hovering, code inserts at the drop line, drop highlight shows then fades out.
- [x] The fade-out is the timeoutEffect path — confirm it still animates and clears.

6. Non-English search — search.worker.ts (I replaced the lunr plugin access with a typed cast — same runtime object)
- [x] Switch UI language and search the reference/API. Cover at least one non-English and Japanese. Confirm results return and are relevant.
- [x] Also spot-check e.g. German/Fr

microbit-matt-hillsdon (approved):

PR #1248, opened 2026-07-29
Adopt @microbit/ui

microbit-matt-hillsdon:
Replaces Chakra UI v2 with microbit UI library based on react-aria-components and Panda CSS.  Style change is pretty minimal with the private brand preset. 

More iteration is planned across the app family to improve theming. This will result in some further widespread change to design tokens. See https://github.com/microbit-foundation/ui/ for some notes.

## Accepted differences

- The button spacing betwen icon and text uses the theme token rather than hardcoded 0.5rem so is affected by the scaling done in the Python whereas it was the only length not affected under Chakra. So ~7px spacing rather than ~8px.
- Sidebar tabs are now a single tab stop (up/down within) which is an accessibility improvement. It's maybe a bit odd how far away "Project" is despite being in the same tab stop.
- Secondary buttons e.g. Save have slightly darker text
- Very minor slider track height change
- Very minor select text position change
- Toast duration forced to >= 5s for accessibility
- Secondary buttons have slightly darker text as part of alignment to the same buttons elsewhere
- Better states for panel expand/collapse buttons
- Hover state for edit project button
- Form fields have 3:1 contrast, e.g. in settings (not looked at slider/switch tracks yet)
- Minor style change in serial hints dialog
- Minor style change in data logging sim table


lang/ui.ca.json
```diff
@@ -1042,6 +1042,22 @@
     "defaultMessage": "Logotip tàctil",
     "description": "Name for the touch logo pin on the V2 micro:bit board. Used in the simulator."
   },
+  "simulator-unit-celsius": {
```
microbit-matt-hillsdon, line 1045: Not an issue, but a reminder to add to Crowdin once merged.

src/common/Spinner.tsx
```diff
@@ -3,19 +3,17 @@
...
     <Box height="100%">
-      <ChakraSpinner
-        display="block"
-        ml="auto"
-        mr="auto"
-        mt={2}
-        label={intl.formatMessage({ id: "loading" })}
+      <UiSpinner
+        css={{ display: "block", ml: "auto", mr: "auto", mt: "2" }}
+        aria-label={intl.formatMessage({ id: "loading" })}
```
microbit-matt-hillsdon, line 16: Isn't this label provided for us?

More generally, can we drop strings from the app and use the ui variants. I guess there's some risk in that.

microbit-matt-hillsdon: No it isn't, I suppose you might occasionally get a more thoughtful label that way.

PR #1258, opened 2026-08-13
  Avoid using aspect-ratio CSS property

microbit-robert:
Not supported in Safari 14. Switch to the AspectRatio component, which
  uses percentage-based padding, and introduce a class that mimics this
  approach for ratios only known at runtime.
  
  See https://github.com/microbit-foundation/ui/issues/72

### DIFF
--- a/src/common/YoutubeVideoEmbed.tsx
+++ b/src/common/YoutubeVideoEmbed.tsx
@@ -3,7 +3,7 @@
  *
  * SPDX-License-Identifier: MIT
  */
-import { styled } from "styled-system/jsx";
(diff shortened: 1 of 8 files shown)


microbit-matt-hillsdon:
This looks fine but holding off on merging to ponder whether there should be any library support for the dynamic aspect ratio thing.

microbit-robert:
> This looks fine but holding off on merging to ponder whether there should be any library support for the dynamic aspect ratio thing.

There is a similar kind of override for the recording graph preview in the settings dialog for ml-trainer, but nothing else in the family needs it.

microbit-matt-hillsdon (approved):

PR #1261, opened 2026-08-13
Adopt @microbit/ui-patterns' LanguageDialog

microbit-matt-hillsdon:
Upgrade @microbit/ui to 0.1.0-alpha.25 and add @microbit/ui-patterns 0.1.0-alpha.2. The dialog becomes a thin wrapper over the shared pattern (flat grid, preview markers, deployment-sourced help-translate link); its hardcoded English preview footnote is finally a translated message. The ui upgrade also brings <html lang> tracking the language setting, via SharedUIProvider.

First-run language now matches browser/OS preferences (after the ?l= hint) via the shared getDefaultLanguageId — previously only ?l= was honoured. Language ids take the registry's canonical BCP 47 casing (zh-CN, zh-TW, es-ES, ga-IE) and the KnownLanguageId union; stored settings and inbound ?l= links with the old lowercase ids keep working via case-insensitive reads, and the lowercase-keyed consumers (catalog imports, typeshed/pyright locale) normalise at their boundaries.

Panda includes the package's sources, compile-lang merges its catalogs, and our help-translate string retires in its favour (`language` stays: the menu item).

To check:
- [x] Language dialog strings themselves are translated as expected
- [x] l= parameter still works
- [x] html lang updates as per chosen default and your choice
- [x] Stored language settings from before still work
    - This is true, but the stored format does change from es-es to es-ES and the old code ignores the new format if we revert. No bad consequences other than the setting falling back to English but we should check other apps don't explode. This came indirectly from wanting a valid code to put on HTML lang and is probably sensible but just need to be careful.


microbit-robert (approved):
LGTM

PR #1287, opened 2026-08-31
Replace framer-motion with a CSS animation

microbit-matt-hillsdon:
framer-motion was added alongside Chakra in the original UI (which required it as a peer dependency) but outlived the Chakra removal to serve a single use: the slide-in transition when navigating the documentation panels. That costs ~35kB gzipped in the main bundle.

Replace it with Panda keyframes applied directly to the panel's VStack, dropping the extra wrapper div. An overshooting cubic-bezier stands in for the spring, and prefers-reduced-motion replaces useReducedMotion. CSS animations leave no transform behind once complete, so the transitionEnd workaround for framer/motion#823 goes too.


microbit-matt-hillsdon:
OK I've stared at the animation until I feel queasy and it seems fine to me.

PR #1292, opened 2026-09-01
Move to ESLint 10 and @microbit/eslint-config 0.2.1

microbit-matt-hillsdon:
Drop the react/display-name carve-out: the rule went with eslint-plugin-react. Worth knowing that a stale carve-out like this does not fail the run, because ESLint tolerates an "off" entry for a rule it cannot find. It just quietly stops doing anything.

eslint-react reports 25 things the old plugin did not. Eight are worth fixing.

Four are leaked conditional renders, all guarding a ReactNode with &&, so a caller passing 0 renders a literal "0". Ternaries with null.

SplitViewSized called createRef in a function component, which builds a fresh ref object on every render. That object is in the effect's dependency array, so the effect re-ran every render. useRef is stable.

Three components seeded state with a function call on every render: InputDialog, project-hooks and router-hooks. Lazy initialisers.

useLast in the language server common module is a Facet combiner, not a hook, and is only used in that file. Now lastOf.

The rest are deliberate and now say so: five hooks hand a useState tuple straight to context, the simulator iframe is same-origin-sandboxed because it is served from a separate origin, XTerm's paste listeners share their lifetime with the terminal DOM as its note explains, Highlight's timeouts are a one-shot scroll and highlight sequence, DocString renders documentation HTML from our own pipeline, search extracts and data log rows are positional, and useDeployment is hook-shaped by design.

npm cannot resolve this bump in one step. The lockfile pins the old shared config with its eslint ^9.7 peer and npm tries to reconcile that rather than replace it, failing with a circular ERESOLVE. Uninstall eslint and @microbit/eslint-config first, then install the new pair.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -33,7 +33,6 @@ export default [
       "@typescript-eslint/no-unsafe-member-access": "off",
       "@typescript-eslint/no-unsafe-return": "off",
       "@typescript-eslint/require-await": "off",
-      "react/display-name": "off",
       // Debt to be reviewed: React-compiler-era hooks rules with existing
       // violations too involved to fix as part of lint config unification.
       "react-hooks/immutability": "off",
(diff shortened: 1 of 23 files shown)


microbit-matt-hillsdon:
LGTM assuming the split view still works.

PR #1298, opened 2026-09-03
Fix drag preview revert wedging further drags (#1296)

### LINKED ISSUE #1296: Drag and drop: revertPreview throws on a stale change set and wedges further drags

Sentry: [PYTHON-EDITOR-NEXT-1H3](https://microbit.sentry.io/issues/5270054960/)

`RangeError: Applying change set to a document with the wrong length`, thrown from `revertPreview` in the drag-and-drop extension.

This is the largest single error source in the editor: **475 events**, 418 of them on v3.1.7 since June. It is not a v3.1.8 regression, but it is the top error by volume on the new release (28 events in the first 19 hours) so it surfaced while reviewing that.

## Cause

`src/editor/codemirror/dnd.ts:96`

```ts
const revertPreview = (view: EditorView) => {
  if (lastDragPos) {
    view.dispatch({
      userEvent: "dnd.cleanup",
      changes: lastDragPos.previewUndo,
      annotations: [Transaction.addToHistory.of(false)],
    });
    lastDragPos = undefined;
  }
};
```

`previewUndo` is a `ChangeSet` captured when the drag preview was inserted. `ChangeSet.apply` throws if the document length has moved on since. So anything that modifies the document between the preview being inserted and the revert — another transaction, or a second drag starting before the first cleaned up — makes the dispatch throw.

The failure is sticky. `lastDragPos = undefined` sits *after* the `dispatch`, so once the dispatch throws, `lastDragPos` is never cleared and every subsequent `dragover` and `drop` re-throws with the same stale change set. That is what the burst pattern looks like: 28 events inside a three-minute window from a single Firefox session on `/v/3/reference/logic`, the last breadcrumb a `code-drop` for `reference-if-elif-else`.

## Stack

```
InputState.handleEvent (@codemirror/view)
  <object>.drop            src/editor/codemirror/dnd.ts:200
    revertPreview          src/editor/codemirror/dnd.ts:98
      EditorView.dispatch
        EditorState.applyTransaction
          ChangeSet.apply  <- throws
```

## Suggested fix

Two things, either of which stops the burst:

1. Clear `lastDragPos` before dispatching, or in a `finally`, so one failure cannot wedge every later drag.
2. Rebase rather than replay — map `previewUndo` through the changes since it was captured, or check `lastDragPos.previewUndo.length === view.state.doc.length` and skip the revert if it no longer matches. Silently skipping leaves the preview text in the document, so mapping is the better of the two if the preview position can be recovered.

Worth adding an e2e case alongside the drag-and-drop preview tests added in 68aac67a: start a drag, mutate the document underneath it, then drop.

microbit-matt-hillsdon:
revertPreview replayed a ChangeSet captured when the preview was inserted. If the document length moved on before the revert, ChangeSet.apply threw, and because lastDragPos was cleared only after the dispatch, every later dragover and drop re-threw with the same stale change set until reload. This is the top error by volume in Sentry.

Three changes, none dependent on which path trips it:

- Clear the stashed drag position before dispatching the revert, so one failure cannot wedge later drags.
- Map the preview undo through any document changes made while the preview is showing, and skip the revert if the length still doesn't match, so an edit under the preview no longer throws.
- Revert a leftover preview when the drag ends by any route, not only via dragleave/drop on the editor. Clearing the drag context (dragend) now reverts the preview and removes the stuck cm-drag-in-progress class, covering a drag cancelled or dropped elsewhere.

Reworked the bare domEventHandlers into a ViewPlugin so the handlers can observe document updates and clean up on view destroy.

Added two e2e cases: a document change during a drag, and a drag that ends without a dragleave. Both fail on the previous code.

Not reproduced manually. The precondition and fix were worked out from the Sentry stack and the code; the mechanism was exercised via emulated drags under Playwright, not a real pointer (Firefox + Chrome).

Fixes #1296

### DIFF
--- a/src/e2e/app.ts
+++ b/src/e2e/app.ts
@@ -569,6 +569,7 @@ export class App {
 
   private dragDataTransfer: JSHandle<DataTransfer> | undefined;
   private dragSource: Locator | undefined;
+  private dragTarget: { clientX: number; clientY: number } | undefined;
 
   /**
    * Starts dragging a code example and hovers it over the target line
@@ -597,10 +598,13 @@ export class App {
       .locator("div")
       .filter({ hasText: targetLine.toString() });
     const target = (await editorLine.boundingBox())!;
-    await this.editorTextArea.dispatchEvent("dragover", {
-      dataTransfer: this.dragDataTransfer,
+    this.dragTarget = {
       clientX: target.x + target.width / 2,
       clientY: target.y + target.height / 2,
+    };
+    await this.editorTextArea.dispatchEvent("dragover", {
+      dataTransfer: this.dragDataTransfer,
+      ...this.dragTarget,
     });
   }
 
@@ -610,12 +614,24 @@ export class App {
     });
   }
 
+  /**
+   * Drops a drag started with startCodeEmbedDrag at the same position.
+   */
+  async dropCodeEmbed() {
+    await this.editorTextArea.dispatchEvent("drop", {
+      dataTransfer: this.dragDataTransfer,
+      ...this.dragTarget,
+    });
+    await this.endDrag();
+  }
+
   async endDrag() {
     await this.dragSource!.dispatchEvent("dragend", {
       dataTransfer: this.dragDataTransfer,
     });
     this.dragDataTransfer = undefined;
     this.dragSource = undefined;
+    this.dragTarget = undefined;
   }
 
   async search(searchText: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
I've spent quite a while trying to figure out a repro for this one that doesn't rely on synthetic drag events without success. The logic holds up but how either of the scenarios identified above happens I do not know. I suspect it's the missing event rather than the doc change which I can't see a source for (at least for length, maybe a diagnostic could arrive late).

microbit-robert (approved):
LGTM. Code changes look good. I can't seem to break anything via drag and drop.

PR #1308, opened 2026-09-07
Adopt @microbit/i18n-tools for translation management

microbit-matt-hillsdon:
We've extended the tooling to support the weird use of Chrome JSON in this project (but with ICU messages) rather than change it in Crowdin, because my testing suggested we'd lose some translations even adding TM pre-translate to the mix, and more importantly we lose all the screenshots that Emma has added over the years to enrich context for translators.

Even though Crowdin JSON is supposed to use a placeholders field it seems like ICU messages are still recognised and provided with editing support in Crowdin, I guess because the format is very generic with just message/description so might be used in other scenarios.


microbit-matt-hillsdon:
Although I've merged this to help test workflow_dispatch, the update-translations.sh script will not work until we deal with similar PRs for stubs and pyright, and it's not really clear whether it's the right approach (vs say, shipping translations with releases of some sort and pulling in translations that way).

PR #1313, opened 2026-09-08
Take the typeshed JSON from @microbit/micropython-microbit-stubs

microbit-matt-hillsdon:
The stubs repository publishes its browser-package output to npm on each release, so the copies in src/micropython/main/ are now generated on install (npm run stubs, part of npm run generate) and gitignored, and Renovate proposes each new release.

update-typeshed.sh and update-translations.sh go with them: each of the three repositories has a weekly translations-download workflow, which is what the combined sync did, and the pyright worker keeps update-pyright.sh.


microbit-matt-hillsdon:
This also brings some very minor fixes.

PR #1321, opened 2026-09-10
Drop structure highlighting runs with a zero-width header

### LINKED ISSUE #1318: [Code structure] skipBodyTrailers passed -1, throwing Invalid position

Sentry: [PYTHON-EDITOR-NEXT-1HJ](https://microbit.sentry.io/issues/5276553570/)

`RangeError: Invalid position -1 in document of length N`, **4,181 events since 29 April 2024**, still present in v3.1.11. This is the largest single group in the project by event count, and unlike the tile crashes in the umbrella issue it is ours, not upstream.

## Cause

`positionsForNode` in `src/editor/codemirror/structure-highlighting/view.ts:116`:

```ts
const bottomPos = skipBodyTrailers(
  state,
  diagnostics,
  end - 1,
  topLineNumber
);
```

`skipBodyTrailers` immediately does `state.doc.lineAt(position)` (`doc-util.ts:35`), which throws for a negative position. So whenever `end` is `0`, we pass `-1` and throw.

`end` is either `block.bodyStart` or `block.bodyEnd`, from the two calls at `view.ts:149` and `view.ts:156`. Both are raw syntax-tree offsets produced by `codeBlocks` in `blocks.ts`, so a node positioned at offset 0 reaches the subtraction unguarded. Lezer error recovery on an incomplete or badly indented document is the plausible source of such a node — the reported document lengths are small (double digits), consistent with mid-typing rather than a finished program.

The `-1` in the message is the tell: it is not a stale-position problem like #1053, it is an unguarded arithmetic edge.

## Suggested fix

Guard at the call site rather than inside `skipBodyTrailers`. Its documented contract takes "a document character position", and `lineAt` throwing on an invalid one is reasonable; the caller is what produces the invalid value. Bailing out (returning `undefined`, as the function already does when it runs past `min`) looks right for `end === 0`, since a block whose body starts at offset 0 has nothing above it to highlight.

Worth a unit test in `doc-util.test.ts` or `view.ts`'s coverage for `end === 0`.

## Related

- #1060 was a `RangeError` in the same `positionsForNode`, but via `view.lineBlockAt` rather than `skipBodyTrailers`, and was closed in 2022.
- #589 is an open report of structure highlighting misbehaving visually on indentation errors. Same error-recovery territory, so a fix here may want checking against that reproduction.

microbit-matt-hillsdon:
A document starting with a colon recovers as a compound statement whose header is a zero-length error node:

```
    ":\n    pass\n"      TryStatement children: ⚠(0,0) Body(0,11)
    "if a:\n    pass\n"  IfStatement children: if(0,2) VariableName(3,4) Body(4,15)
```

start and bodyStart are then both 0, so positionsForNode asked skipBodyTrailers for position -1 and doc.lineAt threw. That accounts for 4181 events in Sentry since April 2024, the largest single group in the project, and it is still present in v3.1.11.

Skip this shape as we cannot meaningfully render it.

Fixes #1318

### DIFF
--- a/src/editor/codemirror/structure-highlighting/blocks.ts
+++ b/src/editor/codemirror/structure-highlighting/blocks.ts
@@ -34,6 +34,7 @@ export interface CodeBlock {
    * Start of the first header node of the run.
    */
   start: number;
+  /** Always past {@link start}, so `bodyStart - 1` is a valid position. */
   bodyStart: number;
   bodyEnd: number;
   /**
@@ -80,13 +81,19 @@ export const codeBlocks = (state: EditorState): CodeBlock[] => {
           let runStart = 0;
           for (let i = 0; i < children.length; ++i) {
             if (children[i].name === "Body") {
-              blocks.push({
-                statement: leaving.name,
-                start: children[runStart].start,
-                bodyStart: children[i].start,
-                bodyEnd: children[i].end,
-                depth,
-              });
+              // Error recovery can give a compound statement a zero-length
+              // error node in place of its header, leaving the body starting
+              // where the header does. There's no header to draw, and the
+              // view relies on bodyStart being past start.
+              if (children[i].start > children[runStart].start) {
+                blocks.push({
+                  statement: leaving.name,
+                  start: children[runStart].start,
+                  bodyStart: children[i].start,
+                  bodyEnd: children[i].end,
+                  depth,
+                });
+              }
               runStart = i + 1;
             }
           }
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
So far as I can tell the only impact of this is the logged errors and stale structure highlighting. Worst case is pasting a repro over a doc with existing structure highlighting.

microbit-grace (approved):
Looks good

PR #1323, opened 2026-09-10
Normalise non-Error values before reporting to Sentry

### LINKED ISSUE #1319: Simulator internal errors reach Sentry as untyped objects

Sentry: [PYTHON-EDITOR-NEXT-3T9](https://microbit.sentry.io/issues/7722464475/)

New in v3.1.11, 3 events. Low priority, but currently untriageable, which is the reason for filing.

## Problem

The group's title is just `Logger.error`, and its exception value is `Error: Object captured as exception with keys: message, name, status`. That is Sentry's fallback when `captureException` is handed something that is not an `Error`, so we get no type, no message and no useful grouping — three events that could be three different faults.

## Cause

```
SimulatorDeviceConnection (src/device/simulator.ts:247)
  Logger.error (src/logging/logger.ts:46)
    reportError (src/logging/sentry.ts:59)
      captureException (@sentry/core)
```

`src/device/simulator.ts:247` handles the simulator's `internal_error` message:

```ts
case "internal_error": {
  const error = event.data.error;
  this.logging.error("Simulator internal error", error);
```

`event.data.error` arrives over `postMessage` from the simulator iframe, so it is a plain object the simulator constructed — `{message, name, status}` — not an `Error` instance. It goes straight to `captureException`, which is why Sentry falls back to the "Object captured as exception" title.

## Suggested fix

Either wrap at the call site in `simulator.ts` so a real `Error` reaches Sentry with the `status` as context, or normalise in `reportError` (`src/logging/sentry.ts`) so any non-`Error` argument is converted and given a meaningful title. The second is probably worth doing regardless, since it protects every other caller of `Logging.error`.

microbit-matt-hillsdon:
The simulator's internal_error message carries Emscripten's ExitStatus, a plain constructor that does not extend Error. After structured cloning over postMessage it is a bare {name, message, status} object, which Sentry titles "Object captured as exception with keys: ..." and groups together regardless of cause.

reportError now converts any non-Error into an Error carrying the original name and message, with remaining fields merged into extra alongside any caller-supplied context. The synthesised stack is dropped because it would only show the logging call chain, identical for every caller, and Sentry would group unrelated errors together on it.

Fixes #1319

### DIFF
--- a/src/logging/sentry.test.ts
+++ b/src/logging/sentry.test.ts
@@ -1,0 +1,90 @@
+/**
+ * (c) 2026, Micro:bit Educational Foundation and contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+import { captureException } from "@sentry/browser";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { reportError } from "./sentry";
+
+vi.mock("@sentry/browser", () => ({
+  addBreadcrumb: vi.fn(),
+  captureException: vi.fn(),
+  init: vi.fn(),
+}));
+
+const dsn = "https://example@sentry.example.com/1";
+
+const captured = () => {
+  expect(captureException).toHaveBeenCalledTimes(1);
+  return vi.mocked(captureException).mock.calls[0];
+};
+
+describe("reportError", () => {
+  beforeEach(() => {
+    vi.spyOn(console, "error").mockImplementation(() => {});
+  });
+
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
It's questionable whether ExitStatus is intended to be reported by this channel. But it's super opaque in Sentry as-is. So let's start with this and confirm what we're actually seeing, then likely dive down a layer into the sim.

PR #1325, opened 2026-09-11
Move the editor's URL state onto react-router

microbit-matt-hillsdon:
Replaces the bespoke pushState/popstate router with react-router 7, as ml-trainer uses, ahead of adding pages beyond the editor. URLs are unchanged: the editor stays at the base URL with the documentation tab and anchor as optional path segments, and deeper paths remain the editor with no tab selected.

- urls.ts holds the basename (derived from the Vite base URL) and the editor route path and link builder, used for both routes and links.
- router.tsx creates the router. The root route wraps its outlet in the existing ErrorBoundary, so an uncaught render error shows the content-load-error message rather than unmounting the whole app.
- router-hooks.tsx keeps the useRouterState/useRouterTabSlug API over useParams/useNavigate/useLocation. Focus travels in history state and the state identity is tied to location.key so navigating to the same anchor again re-runs scroll and focus effects, as before.
- Search results use react-router's Link so the href respects the basename (the old toUrl produced a broken absolute URL).
- The sidebar only resets the anchor when the clicked tab is the current one. It previously reset whenever an anchor was set and relied on the old router having updated state synchronously between pointer-up and click; with react-router that made clicking another tab from a deep link land on the current tab's top level.

### DIFF
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -1,0 +1,31 @@
+/**
+ * (c) 2026, Micro:bit Educational Foundation and contributors
+ *
+ * SPDX-License-Identifier: MIT
+ */
+import { createBrowserRouter, Outlet } from "react-router";
+import ErrorBoundary from "./common/ErrorBoundary";
+import { basename, editorRoutePath } from "./urls";
+import Workbench from "./workbench/Workbench";
+
+export const createRouter = () =>
+  createBrowserRouter(
+    [
+      {
+        id: "root",
+        path: "",
+        // Without this an uncaught render error unmounts the whole app.
+        element: (
+          <ErrorBoundary>
+            <Outlet />
+          </ErrorBoundary>
+        ),
+        children: [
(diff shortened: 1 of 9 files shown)


src/router.tsx
```diff
@@ -0,0 +1,31 @@
...
+ */
+import { createBrowserRouter, Outlet } from "react-router";
+import ErrorBoundary from "./common/ErrorBoundary";
+import { basename, editorRoutePath } from "./urls";
+import Workbench from "./workbench/Workbench";
+
+export const createRouter = () =>
+  createBrowserRouter(
+    [
+      {
```
microbit-matt-hillsdon, line 14: You might expect 404 handling but this matches current behaviours.

microbit-matt-hillsdon:
I'm going to merge this to help unblock some other work but @microbit-robert please can you review retrospectively.

PR #1326, opened 2026-09-11
Show the shared error page for uncaught render errors

microbit-matt-hillsdon:
Takes @microbit/ui-patterns 0.6.1 and uses its ErrorBoundary and
UnexpectedErrorPage at the root route. The react-router change had put
the app's own ErrorBoundary there, whose inline "download your hex
file" text assumes the editor is still on screen; full-page it is not.
That boundary is unchanged around the sidebar panels, where a
documentation tab failing to load (typically a blocked Sanity domain)
still shows the inline message beside a working editor.

The page shows an error reference so a support request can be matched
to the Sentry report. Logging.error now returns that reference (the
event id from captureException) when the error was sent, and the root
layout passes it through to the page. The support link is the brand's;
deployments without one get no support sentence, as with the help menu.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@microbit/microbit-fs": "^0.10.0",
         "@microbit/micropython-microbit-stubs": "^1.0.0",
         "@microbit/ui": "^0.4.0",
-        "@microbit/ui-patterns": "^0.5.0",
+        "@microbit/ui-patterns": "^0.6.1",
         "@sanity/block-content-to-react": "^3.0.0",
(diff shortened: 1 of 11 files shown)


microbit-matt-hillsdon:
Evil hack to test:

```
(() => {
  const orig = Element.prototype.querySelector;
  Element.prototype.querySelector = function (sel) {
    if (String(sel).includes("data-panel-content")) {
      throw new Error("review: forced render error");
    }
    return orig.call(this, sel);
  };
})();
```

Then change tab. Will merge and re-run on beta which should have the actual Sentry reference.

PR #1327, opened 2026-09-11
Multiple projects storage foundation

microbit-matt-hillsdon:
Note: targets new home-page integration branch as part of a multi-step plan.

The file system is now mirrored to IndexedDB instead of session storage.
- `projects-db.ts`: two stores, project metadata and files keyed by `[projectId, name]`, so listing is cheap and a change writes only the files it touched. Opening asserts the stores exist and throws `VersionError` otherwise.
- `indexeddb-storage.ts`: an `FSStorage` for one project, used as the `SplitStrategyStorage` secondary. Keystroke writes coalesce per file and flush as one transaction after 300ms and when the page is hidden. A failed flush is reported, not thrown, and its changes dropped: the in-memory primary still has them and retrying against a full quota helps nobody.
- `current-project.ts`: which project a tab opens. The current id lives in session storage, so a project is still per tab. A session-storage project is migrated on first load so a reload after deploy lands in the user's work; only the `fs/` keys are removed, since session settings share that storage. Without IndexedDB, or with an incompatible database on a public stage, the editor falls back to session storage as before.
- `SplitStrategyStorage` accepts a promise for its secondary, since opening IndexedDB is asynchronous and the Host API is not.

Production and beta get separate databases by base path (`python-editor/v/3`, `python-editor/v/beta`). Review builds share `python-editor-review`: they are internal and a project from one branch is useful on the next. If/when we hit a schema change, non-public stages show a "Breaking change to stored data" page with clear-and-reload, which is useful during development.

Adds `FileSystem.switchStorage`. Opening a project will swap the file system's backing storage rather than replace files one by one. Waits for any in-flight initialise, makes the new storage the record, refills the hex file system, bumps every file's version so a same-named file reloads in the editor, and notifies. Not yet called outside tests.

## User-visible behaviour on this branch

The difference is that a project now survives the tab closing. Nothing lists the projects yet. This is obviously weird hence the integration branch, but a step in the right direction!

## Known issues, to be resolved by later plan steps

- Every new tab creates a project, so the library accumulates "Untitled" projects until we deal with reopening.
- A `#project:` link (microbit.org "open in Python Editor") still replaces the current project's files, as it does today. Later work will make it create a new project.
- Two tabs on the same project do not sync; last flush wins.
- No quota toast yet: a failed flush is logged through `Logging.error`. The toast comes with the pages.


microbit-robert (approved):
LGTM

PR #1330, opened 2026-09-14
Projects pages

microbit-matt-hillsdon:


microbit-robert:
It's a lot of change, but I think this looks good. Have not tested the iframe / no indexedDB version.

microbit-matt-hillsdon:
> It's a lot of change, but I think this looks good. Have not tested the iframe / no indexedDB version.

I think it's fine for us to review that separately and follow up with fixes on home-page if needed.

PR #1339, opened 2026-09-17
Polyfill Array.prototype.at mainly for Safari < 15.4

microbit-matt-hillsdon:
marked 18 calls `.at` while parsing, which is ES2022 and so needs Chrome 92, Safari 15.4 or Firefox 90. Our floor is chrome90/firefox88/safari14.1/ios14.5 (BUILD_TARGETS in vite.config.ts, mirrored in browserslist), and esbuild lowers syntax but never built-ins, so Safari and iOS 14.1-15.3 users get a TypeError from the lexer. The error boundary catches it, so the whole docs panel is replaced by the error page rather than the markdown degrading.

Reported once from the wild against v3.1.13
(PYTHON-EDITOR-NEXT-3V7) from a browser below the floor, but the same call fails on Safari versions we do claim to support.

tsconfig's ES2021 lib already stops our own source calling `.at`; this is only needed because the call is in a dependency.

Prime candidate for removing on next floor bump.


src/polyfills.ts
```diff
@@ -0,0 +1,28 @@
...
+// enough to run them, so the arithmetic would otherwise never be exercised.
+export function at<T>(this: ArrayLike<T>, index: number): T | undefined {
+  const i = Math.trunc(index) || 0;
+  return this[i < 0 ? this.length + i : i];
+}
+
+for (const proto of [Array.prototype, String.prototype]) {
+  if (!(proto as { at?: unknown }).at) {
+    // Non-enumerable, matching the native method: a plain assignment would
+    // show up in `for...in` over an array.
```
microbit-matt-hillsdon:
```suggestion
```

src/polyfills.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ */
+// `Array.prototype.at` is ES2022: Chrome 92, Safari 15.4, Firefox 90. That is
+// above the support floor in vite.config.ts, and esbuild lowers syntax but
+// never built-ins, so anything below the floor needs this shim. marked calls
+// it while parsing, which takes out the docs panel and language-server hovers
+// on affected browsers rather than degrading quietly.
+// Exported for tests: the install below is a no-op in any runtime new
+// enough to run them, so the arithmetic would otherwise never be exercised.
```
microbit-matt-hillsdon:
```suggestion
// Remove when we drop Safari < 15.4.
// Also seen helping some v old Chrome but not reasonable to maintain support here past when we drop older Safari.
```

PR #1341, opened 2026-09-18
Align npm scripts and dev ports

microbit-matt-hillsdon:


.github/workflows/build.yml
```diff
@@ -60,10 +60,6 @@ jobs:
       - run: npm run ci
         env:
           VITE_SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
-      - run: mkdir -p /tmp/app${BASE_URL} && cp -r build/* /tmp/app${BASE_URL} && npx serve --no-clipboard -l 3000 /tmp/app &
```
microbit-matt-hillsdon, line 63: It is more than a little amusing that we've not been using this for some time.

